PR #7, opened 2023-11-29
fix: `navigator.mediaDevices` returning `undefined` on macOS

mooyg:
Fixes the issue mentioned [here](https://github.com/cap-so/cap/issues/5#issue-2014901879) through PLIST files to get the permissions for the media devices.
Another fix for encoder being registered twiced.

This might still not work on LINUX as the tauri linux webview doesn't support it.


richiemcilroy:
Thanks for this @mooyg - merged!

PR #10, opened 2023-12-08
fix: ffmpeg not found

mooyg:
* Uses which to find the path for ffmpeg
* Uses absolute path for recording directory too

### DIFF
--- a/apps/desktop/src-tauri/src/main.rs
+++ b/apps/desktop/src-tauri/src/main.rs
@@ -13,6 +13,7 @@ use ffmpeg_sidecar::{
 
 mod recorder;
 use recorder::ScreenRecorder;
+use which::which;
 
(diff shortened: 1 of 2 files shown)


richiemcilroy:
Hey! Thanks for the submission. Tried running but I get this error:

`desktop:dev: Thread panicked: PanicInfo { payload: Any { .. }, message: Some(called `Result::unwrap()` on an `Err` value: CannotFindBinaryPath), location: Location { file: "src/main.rs", line: 81, col: 54 }, can_unwind: true }`

The existing code in main works for me.

What device are you using?

mooyg:
Oh that's weird, do you have ffmpeg in the path? I didn't do error handling there. I am on macos 14.1.2 (23B92)

Issue #17, opened 2024-03-23
Corrupt frame data after screen capture (macOS)


richiemcilroy:
**Description**
Experiencing a bug with frame data (from screen capture) being corrupt on some Mac's. Seems to work fine on my 16" MacBook Pro, but when testing on 14" MacBook Pro's the frame image looks corrupt.

**My thoughts**
I believe this possibly stems from the stride calculation needing to be altered. This is because the frame data *can* be visible if you mess around with the stride settings - it's not always completely corrupt. The pixel data can be shown, it's just a matter of how it is represented. Also, it may be to do with the scaling factor of the device.

**Example of a corrupted frame**
<img width="1599" alt="Screenshot 2024-03-23 at 02 48 50" src="https://github.com/CapSoftware/cap/assets/33632126/71bfe58b-accb-4e04-b073-f872300b9178">

**Device Specs**
14" MacBook Pro
M1 Pro Chip
Sonoma 14.4 (also tested on early versions, e.g. 13.1)

**Resolutions tested**
1800x1169 (also 2x scale)
1512x982 (also 2x scale)
1352x878 (also 2x scale)

**Suspect code**
The code for this is located at `media.rs` which belongs to `/apps/desktop`: https://github.com/CapSoftware/cap/blob/main/apps/desktop/src-tauri/src/media.rs

The line you're looking for is `line 356`, `match capturer.frame() {`. In here, we are iterating over each row and extending the slice to send to `video_channel_sender`. This then feeds into the video ffmpeg command, at `line 437`. 

The raw frame should be packed bgra (ARGB8888). If you search for `ARGB8888` in the codebase, you'll find the `capturer` package, which ultimately creates the core graphics display stream via `CGDisplayStreamCreateWithDispatchQueue`.

shamun-khatri:
@richiemcilroy I want to work on this but is this only related to 14" MacBook Pro and are other users also facing the same prob@richil em on any other models?

richiemcilroy:
Hey! I can't be sure that it's definitely only 14". It seems to be on retina devices, but for some reason not my 16" 🤔 so it's pretty strange. > @richiemcilroy I want to work on this but is this only related to 14" MacBook Pro and are other users also facing the same prob@richil em on any other models?

okhot:
@richiemcilroy is there anyway to reproduce this?

richiemcilroy:
> @richiemcilroy is there anyway to reproduce this? 

[...]

PR #18, opened 2024-03-24
Prevent frame data corruption on certain Mac configurations

filleduchaos:
This addresses the bug raised in #17 which is caused by a performance optimisation in macOS's Core Graphics that can insert padding into pixel buffers to keep them optimally aligned for the hardware in use.

I've added comments further explaining the problem and its solution, but as some additional info I'm fairly certain that this is only affecting 14-inch and not 16-inch MBPs because they have native screen widths of 3024 and 3456 pixels respectively. While the exact hardware alignment isn't public info AFAIK, the 16-inch Pro's width is cleanly divisible by powers of 2 up to 128 while the 14-inch Pro's width runs into fractions from 32 up. I don't have a 14 inch Pro on hand so I tried to find a similarly odd external monitor to confirm my suspicions with, but Apple has apparently truly outdone themselves with this resolution 😅 

I did also clean up the video capturing code a bit because it was a little difficult to follow, but I didn't change much there.

/claim #17

### DIFF
--- a/apps/desktop/src-tauri/src/capture/src/common/dxgi.rs
+++ b/apps/desktop/src-tauri/src/capture/src/common/dxgi.rs
@@ -38,6 +38,13 @@ impl Capturer {
 
 pub struct Frame<'a>(&'a [u8]);
 
+impl Frame <'_> {
+    pub fn stride_override(&self) -> Option<usize> {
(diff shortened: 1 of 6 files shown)


richiemcilroy:
Amazing work!

Issue #28, opened 2024-04-26
Cap.so Raycast extension


simonayotte:
**Support Cap.so with an official Raycast extension.**

**Includes minimally** two commands that can be invoked from Raycast:
- Start recording (w/ default settings set by user)
- Stop recording (Copies the video URL by default at the end of the recording)

Ideally includes support for:
- Opening the web dashboard

neo773:
Could I get assigned to this?
I’ve already authored 3 extensions on [Raycast](https://www.raycast.com/huzef44)

hknhhy333:
good

richiemcilroy:
This is currently in progress by @ItsEeleeya - almost finished getting it released!

richiemcilroy:
Sorry for the slow progress here. Over the last couple of months we have completely redeveloped the app from the ground up, leading to the launch of Cap v0.3.0. It's a different architecture (Tauri v2 + SolidJS).

@ItsEeleeya - would you like to look at updating for the new version of the app?

ItsEeleeya:
Definitely, I've only been watching the rapid re-make of the app, it's looking awesome btw!
I'm looking to add more features to this soon and then re-implement it for the app.
One of the features I've had in mind was connecting your cap account to the extension so you could view/edit/share any recordings or screenshots.
Raycast only supports PKCE for OAuth. I believe NextAuth supports this method but I might need a bit of help making extra sure this part is implemented correctly!

It think it should also allow you to only use the web recorder and trigger the start/stop via the extension.

I'll be mostly available by the end of the next weekend!

raphaelbernhart:
Just checking in, is this still needed? :)

iamsoum:
Is this still being worked on? I don't see any extension in Raycast and the this issue is not closed yet.

ItsEeleeya:
@raphaelbernhart @iamsoum Yes, it's going to be updated soon for the new version of Cap
I've been working on the Windows compatibility recently and this will be my priority

svliantiss:
Is this working yet, I'd love to use it :)

JYZ-LESLIE:
/attempt #28

Scoped plan for current v0.3 architecture:
1. Add lightweight  and  deep links in desktop so external tools can trigger start/stop without requiring capture-mode payloads.
2. Make  start using the user’s existing recording defaults (target/mic/camera/mode/system-audio) from .
[...]

PR #38, opened 2024-05-05
macOS Raycast Extension

ItsEeleeya:
Solves #28 

Added deep-link support.
The Raycast extension will be able to use these to send commands to Cap.
Currently, you can start/stop recording with specified media devices, as well as open your account's Dashboard.

I'm not sure if this will interfere with the auth process. This should be tested.

To test on macOS, make sure you first build a debug version of the app via `pnpm tauri:build --debug` and then "install" it. (Putting the resulted `.app` into your `Applications` folder.
_Read more about it [here](https://github.com/FabianLars/tauri-plugin-deep-link/tree/main?tab=readme-ov-file#macos)_
Then run the app via `pnpm dev`

The structure of the deep-link scheme is like `caprecorder://:command?parameters`

Copy and paste the URI below into a browser to try it out:
`caprecorder://start-recording?mic_in_label=MacBook%20Pro%20Microphone&vid_in_label=FaceTime%20HD%20Camera`

The URI must be encoded. You can also use `none` as the value for a media device label to disable it.

P.s: I'm not sure where to create the extension. Should it be under the packages directory?

### DIFF
--- a/apps/web/app/api/video/title/route.ts
+++ b/apps/web/app/api/video/title/route.ts
@@ -4,8 +4,10 @@ import { nanoId } from "@cap/database/helpers";
 import { videos } from "@cap/database/schema";
 import { db } from "@cap/database";
 import { eq } from "drizzle-orm";
+import { headers } from "next/headers";
+import { rateLimitMiddleware } from "@/utils/helpers";
 
-export async function PUT(request: NextRequest) {
+export async function handlePut(request: NextRequest) {
   const user = await getCurrentUser();
   const { title, videoId } = await request.json();
   const userId = user?.id as string;
@@ -59,3 +61,8 @@ export async function PUT(request: NextRequest) {
     })
   );
 }
+
+export const PUT = (request: NextRequest) => {
+  const headersList = headers();
+  return rateLimitMiddleware(10, handlePut(request), headersList);
+};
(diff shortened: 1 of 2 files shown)


ItsEeleeya:
The extension is [here](https://github.com/ItsEeleeya/Cap-Raycast-Extension)

Screenshot showing Raycast open with the "Start Recording" command from Cap selected.

To test, make sure Cap is built with `--debug` and installed. Then it has to be running in development.
For the extension, just `npm install` & `npm run dev`

ItsEeleeya:
There will be a Form view that opens up once you invoke the Start Recording command, letting you select the input devices and then start your Cap.
Getting the device labels probably has to be implemented in Swift.

ItsEeleeya:
Updated the [extension](https://github.com/ItsEeleeya/Cap-Raycast-Extension/tree/main) to allow for selecting the devices to record as seen below:

Screenshot showing Raycast open with the "Start Recording" command showing a form to select a microphone and camera to launch Cap with.

_Tested with the commit [84c54a660933f8f87f277a46c4a04eebf4b96729](https://github.com/CapSoftware/Cap/commit/84c54a660933f8f87f277a46c4a04eebf4b96729)._

Note:
 I'm not sure if this can happen, but both Cap and the extension use localized names for identifying the device. There could be a chance that having a different system language might result in a mismatch or cause either to fail.

ItsEeleeya:
Hey @richiemcilroy, could you please have a look at this?
_I'd appreciate it if you could also check your Twitter messages!_

PR #45, opened 2024-06-12
Use Cargo workspace

Brendonovich:
_Don't review until #44 is merged_

Adds a root `Cargo.toml` that turns the repo into a Cargo workspace, and moves `capture` into `crates/capture`, a more common method of organising workspace crates.


richiemcilroy:
Trying to get this one merged, but for some reason it breaks opening the Camera window :(

Brendonovich:
I think that was due to a bunch of dependencies updating, I've redone it to make sure the Cargo.lock is reused

PR #46, opened 2024-06-13
Combine video and audio into one ffmpeg process

### LINKED ISSUE #24: Combine Ffmpeg commands into one single command (macOS)
Right now, there are two separate commands for capturing audio and video. They can be found in `media.rs` located in `apps/desktop/src-tauri`. I would like to have them combined into one single ffmpeg command to reduce system load when recording.

**Requirements**

- Create one single ffmpeg command that takes input from both the audio and video sources, muxing into one .ts file
    - This will probably be done by creating separate named pipes as the input sources
- Make sure the recording process is still performant after changing

You can use [this guide](https://github.com/CapSoftware/Cap/blob/main/CONTRIBUTING.md#how-do-i-run-the-desktop-app-locally-without-needing-to-use-auth) to help you get started developing locally as quickly as possible.

Brendonovich:
_Don't review until #45 is merged_

Combines the audio and video ffmpeg arguments into a single command that receive each stream via named pipes.
Video segments are now uploaded to `${userId}/${videoId}/segment/`, and the default audio selector is always enabled for MediaConvert. Not sure if that's all been done properly, but the ffmpeg stuff in `media.rs` should be good.

/claim #24
Closes #24


richiemcilroy:
This is awesome @Brendonovich, great work!

PR #48, opened 2024-07-04
Implement deeplink commands for the Raycast Extension

ItsEeleeya:
Solves [28](https://github.com/CapSoftware/Cap/issues/28)

Added deep-link command support.
The Raycast extension will be able to use these to send commands to Cap.
Currently, you can start/stop recording with specified media devices, as well as open your account's Dashboard.

What needs testing:
 Make sure OAuth still works.

To test on macOS, make sure you first build a debug version of the app via `pnpm tauri:build --debug` and then "install" it. (Putting the resulted .app into your Applications folder.)
_Read more about it [here](https://github.com/FabianLars/tauri-plugin-deep-link/tree/main?tab=readme-ov-file#macos)_
You can find the dmg/app at the path below:
`<repo>/apps/desktop/src-tauri/target/universal-apple-darwin/debug/bundle/[dmg | macos]`

Then run the app via `pnpm dev`

The structure of the deep-link scheme is as `caprecorder://:command?parameters`

Copy and paste the URI below into a browser to try it out:
`caprecorder://start-recording?mic_in_label=MacBook%20Pro%20Microphone&vid_in_label=FaceTime%20HD%20Camera`

The URI must be encoded. You can also use `none` as the value for a media device label to disable it.

You can find the source for the extension [here](https://github.com/ItsEeleeya/Cap-Raycast-Extension)

To test the extension, make sure Cap is built with `--debug` and installed. Then run it in development.
Then in the extension directory, run `npm install` & `npm run dev`

Screenshot showing Raycast open with the "Start Recording" command showing a form to select a microphone and camera to launch Cap with.


ItsEeleeya:
> Looks great, got it all working @ItsEeleeya! is it possible to remember the previous mic/webcam state in Raycast when you want to record? or maybe pull the current selected state directly from Cap?
> 
> Other than that, think we're ready to go

Yes, I believe Raycast has a preserve option for forms and the name of devices can be stored.
The potential downside is if you've removed the selected device since the last time you used the extension, then there could be a slight delay in loading device labels. 
(I'm not sure how much faster the built version of the swift code is, with debug it takes a couple of seconds for me)

About pulling the state from Cap, if it's done with deeplinks both apps might flash in and out for a split second so I'm not so sure about it yet.

ItsEeleeya:
Done! Now it caches the devices labels and your selected device. 
Since it's using Raycast's default behavior, you need to perform the primary action on the form before it stores the selected devices.
It still shows the loading indicator as it tries to resolve device name.

The changes are [here](https://github.com/ItsEeleeya/Cap-Raycast-Extension/commit/4cf413176acb0129308a29a197173c104874e612)

ItsEeleeya:
Closing this for now. Will add more features on the extension and implement this again 👀

PR #51, opened 2024-07-16
Use `@mattrax/mysql-planetscale` in dev

Brendonovich:
Implements the `@mattrax/mysql-planetscale` library in development to allow `@planetscale/database` to connect with a plain mysql database.
The way i've implemented this requires moving from the `DB_*` environment variables to a single `DATABASE_URL` connection string.
The url needs to start `mysql://` during development and when using drizzle kit, and should start with `https://` in production or when testing locally using planetscale.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,7 +26,12 @@ NEXT_PUBLIC_TASKS_URL=http://localhost:3002
 NEXT_PUBLIC_LOCAL_MODE=false
 
 # -- database ****************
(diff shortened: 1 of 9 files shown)


richiemcilroy:
Thanks for this @Brendonovich! much appreciated. Also added in a new `docker-compose.yml` file, and updated the `dev` to include the creation of the DB + migration etc.

Issue #25, opened 2024-04-11
Make camera overlay window more stable (macOS)


richiemcilroy:
**Issue**
It's possible to break the camera window (window turns black) when the app is first launched. Seems to happen if the app has launched in the background, then you open it after the camera preview has already initialised.

![Screenshot 2024-04-11 at 14 00 51](https://github.com/CapSoftware/Cap/assets/33632126/c30cab4a-b998-466b-a9df-ff4b117d010c)

You can use [this guide](https://github.com/CapSoftware/Cap/blob/main/CONTRIBUTING.md#how-do-i-run-the-desktop-app-locally-without-needing-to-use-auth) to help you get started developing locally as quickly as possible.

richiemcilroy:
Hey @feliciien are you working on this?

feliciien:
yes

Le sam. 13 avr. 2024 à 12:30, Richie McIlroy ***@***.***> a
écrit :

richiemcilroy:
/bounty $100

Upped the bounty here, would love to get this one fixed

richiemcilroy:
Upped the bounty here, would love to get this one fixed

Vijaykv5:
Giving it a try @richiemcilroy 
/attempt https://github.com/CapSoftware/Cap/issues/25

richiemcilroy:
awesome, thank you @Vijaykv5! 

You can refer to this section here in the contributing guide: 

https://github.com/CapSoftware/Cap/blob/main/CONTRIBUTING.md#how-do-i-run-the-desktop-app-locally-without-needing-to-use-auth

ItsEeleeya:
Is the previous message automated?
I cannot reproduce the issue even when I set the app to be unfocused and minimized on start. It happens at random when I change the input video quickly between 3 devices.

Also, I keep getting the update available message even though my fork is synced with the main branch.

ItsEeleeya:
Hey @richiemcilroy! Could you please try out [this branch](https://github.com/ItsEeleeya/Cap/tree/fix-camera-overlay-macos) and check if it's solved the issue? 
If not I'd greatly appreciate it if you could provide a little more info on how to reproduce this.

Also, I came across a few other issues trying this. It seems like switching between cameras doesn't work properly and it gets stuck on the same input the app started with (Same commit, and the latest release).

[...]

PR #56, opened 2024-07-21
(try) Fix camera overlay

ItsEeleeya:
Might be a fix for #25 
- fix: closing overlay now properly sets the selected video to none.


richiemcilroy:
Looks good @ItsEeleeya thank you.

An easy way to replicate this issue, is once the webcam has loaded, right click the main window and reload. This will then cause the webcam to become disabled and it will no longer work.

Screenshot 2024-07-21 at 17 31 36

ItsEeleeya:
Hey @richiemcilroy, could you please try this again?

From what I saw, after reloading the main window, trying to use `WebviewWindow.getByLabel` would return null even with the window existing. It seems to work when called on the backend so the camera overlay should now properly close and re-initialize.

richiemcilroy:
Seems to work a lot better now @ItsEeleeya!

richiemcilroy:
Thank you :)

ItsEeleeya:
oh, that's great! no problem

By the way, do you also have an issue where no matter what video input you select, the overlay just shows the same input (seems to be the first index)?
I was also going to add an option to preserve the last selected device.

richiemcilroy:
ahh yeah, I see what you mean... hmm, would be good to get that fixed. Feel free to throw up another PR if you manage to

richiemcilroy:
also, I think you need to add a claim to this so I can reward the bounty

ItsEeleeya:
> also, I think you need to add a claim to this so I can reward the bounty

Could you please check your Twitter/X for that? I have messaged you

PR #61, opened 2024-07-26
Record.tsx refactor

Brendonovich:
This PR moves around all the web recorder logic into different files and functions with the goal of untying as much as possible from React.

`useRecorder.tsx` contains all the low level media processing logic including muxing and uploading, and exports `useRecorder` for easy creation and control of the recorder. `@tanstack/store` is used to track the recorder state, as this can be easily switched to other frameworks with their respective adapters.

Devices and media streams are now managed by `@tanstack/react-query`. This makes the fetching of the actual devices more declarative, as whenever `*DeviceId` change the query will re-run.

Most of the shared logic used for controlling the `<Rnd>` elements has been moved to `usePreviewRnd`, so that resize logic can be handled in one place.


richiemcilroy:
unreal, so much nicer. Works great.

PR #68, opened 2024-08-02
Use client's stream.m3u8 when generating playlist dynamically

Brendonovich:
`/api/playlist` currently lists all media files to construct the playlist. This is slow and error-prone:
1. S3 commands take a while, especially in local dev for me in Australia
2. When `ListObjectsV2Command` is executed, the client may not be done uploading all its files.

I've changed `/api/playlist` to pull the uploaded `stream.m3u8` which is basically guaranteed to exist, and modify all the `segment.ts` entries to be presigned urls. Some of the later segments may not exist when the client goes to load them, but this is something the client needs to be able to handle anyway. Reading one file from S3 rather than making a bunch of requests shaves finish recording -> view in share page time from 8s to 5s for me.

### DIFF
--- a/apps/web/app/api/playlist/route.ts
+++ b/apps/web/app/api/playlist/route.ts
@@ -90,48 +90,35 @@ export async function GET(request: NextRequest) {
     });
 
     if (video.source.type === "local") {
-      const objects = await s3Client.send(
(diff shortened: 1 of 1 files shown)


richiemcilroy:
Works great!

PR #71, opened 2024-08-03
Migrate to Tauri v2

ItsEeleeya:
**This is currently a WORK-IN-PROGRESS.**
With the recent release of Tauri 2.0.0-rc, I decided to give this another try after a couple of weeks since my last attempt.
Most things should work as expected.

#### What's currently missing:
- [x] All dependencies updated for v2
- [ ] Tray stop icon when recording is in-progress.
- [ ] _Deep-links and re-implementation of_ #48 
- [ ] Camera overlay (Not working)
- [ ] Fix any missing permissions

#### New changes:
* Using [tauri-plugin-decorum](github.com/clearlysid/tauri-plugin-decorum/) for webview transparency and native traffic lights offset.
This forces using default radius on the window. We either can choose between keeping a custom radius or use Tao's default implementation which requires macOSPrivateAPIs. I'd appreciate any feedback on this for now.

#### New coming changes:
* (macOS) Show timer on the tray icon when recording is in-progress.
* (macOS) Add more info to the `About Cap` window.
* (Windows/Linux) Optimize the titlebar to better suit other platforms.
* _And more…_

Any feedbacks or requests are appreciated!


ItsEeleeya:
[7a55c41](https://github.com/CapSoftware/Cap/pull/71/commits/7a55c41d28a04ddb433126262e45a534f3cd22d1) preview:


The icon on the right is from Cap. (For me, it seems to update every two seconds or so which might be from Tauri, I'm not sure)
The text looks a bit odd because it's using Unicode monospace characters instead. Otherwise the width of the menu item would change and cause the rest of the items to shift as well.

I'm not sure if this should be kept. To make it look more like the one on the left (From the Clock app) I might need to make a plugin to allow for the creation of [NSStatusItem].(https://developer.apple.com/documentation/appkit/nsstatusbar?language=objc)
We might be able to  also make it look like a chip component, or the green camera indicator.

ItsEeleeya:
If anyone testing this experiences this same issue, or has more info on them please let me know!
I seem to get these frontend errors randomly. Only on first launch.
Main window (unexpected EOF):
sc1

Camera window (Hydration error):
sc2

ItsEeleeya:
There are a few more changes in progress, but this should be ready for a review.

Brendonovich:
Most things seem to work properly, just a couple of things I've ran across:
The version of `tauri-plugin-oauth` might need to be updated in `Cargo.lock`, the currently locked version didn't compile with tauri rc 0 on my machine.
Styles are a bit off in dark mode, not sure if this is a tauri 2 thing but it doesn't happen on `main`
image

ItsEeleeya:
Regarding the styles, I was thinking wondering the same thing. It's happened for my previous PRs too even though I've made no changes to them and the branches are based on `main`

ItsEeleeya:
I just built directly from main and it looks like this which it always did for me (at first I thought it was intentional)

ItsEeleeya:
Hey @Brendonovich, this is now ready for a review!
The previous issue regarding the look was just due to the app missing styles in dev mode.
Make sure to test the recording and check if you also get any random hydration errors

Brendonovich (approved):
Seems to work great now 👍

ItsEeleeya:
Hey @Brendonovich, I'm so sorry I forgot one of the permissions!! It's for plugin-oauth.
I only realized this when I ran the whole project with Docker and couldn't log in

Brendonovich:
Ah thanks for mentioning that, I'll add it back in!

PR #75, opened 2024-08-16
feat: upload speed check & test record

### LINKED ISSUE #73: Implement speed test and health check for Cap app
Need to implement the following functionality into Cap:

- Speed test: A small indicator in the bottom left or right of the Cap desktop app which shows an indication of the current upload speed. Depending on the upload speed, we should be using different quality types in the recording process. Right now we are using `Resolution::Captured`, but we should be using a smaller size depending on the current upload seed.

- Health check: A function which runs on startup to test whether or not uploading is possible. It will attempt to upload a screen recording test to an endpoint created via the Web App (API route). Depending on the outcome, show an indicator in the Cap Desktop app to let the user know that something went wrong and to contact support. This might need to be a little bit more defined but we can discuss this. 

This will allow us to be able to control the quality of the outcome of Cap's currently being recorded, and have less frustration when a user tries to record a Cap and it ultimately failed because of upload speed, or a recording issue.

Would love to work closely with whoever takes on this task - looking to get it implemented ASAP!

neo773:
This PR adds:

1. Network speed check: Measures upload speed to set video quality.
uses macOS builtin `networkQuality` for private and secure speed tests.

2. Test recording: Performs a brief recording to validate setup during onboarding screen.
Validates the segment and random frames.

/claim #73
/closes #73


neo773:
@Brendonovich

This is ready for review

https://github.com/user-attachments/assets/14f69979-0f4b-4595-8127-406b66c5da96

PR #81, opened 2024-09-05
Add CrabNebula Release CI for `desktop-solid`

atilafassina:
This PR creates a `release.yaml` workflow for the new app (`desktop-solid`).
It also adds setup for the CrabNebula Auto-Updater and uses the same versioning system as the previous one.

---

I also added a  release profile, this will likely reduce the build time and the bundle size of the production app by removing some debugging code and some links for better stack traces.

ping @lucasfernog for a sanity check 😉


.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Get app version
+        id: get_version
+        run: echo "appVersion=$(jq -r '.version' apps/desktop/package.json)" >> $GITHUB_OUTPUT
+
+      - name: create daft release
```
lucasfernog-crabnebula:
```suggestion
      - name: create draft release
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+      contents: write
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Get app version
+        id: get_version
+        run: echo "appVersion=$(jq -r '.version' apps/desktop/package.json)" >> $GITHUB_OUTPUT
```
lucasfernog-crabnebula, line 21: we should probably pull the version from tauri.conf.json instead (that's also the first option that Tauri uses if it is set). Otherwise just remove it from tauri.conf.json so they both use the same source of truth - version mismatches might lead to the asset not being found

.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Create .env file in root
+        run: |
+          echo "appVersion=${{ steps.get_version.outputs.appVersion }}" >> .env
+          echo "CAP_DESKTOP_SENTRY_URL=https://efd3156d9c0a8a49bee3ee675bec80d8@o4506859771527168.ingest.us.sentry.io/4506859844403200" >> .env
+
+      - name: Copy .env to apps/desktop-solid
+        run: cp .env apps/desktop-solid/.env
+
+      - name: Output .env file
```
lucasfernog-crabnebula, line 59: this is a debug step right?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Build MacOS Apps
+        # Install x86_64-apple-darwin for mac and build Tauri binaries
+        # Build both intel and apple silicon
+        working-directory: apps/desktop-solid
+        run: |
+          rustup target add x86_64-apple-darwin 
+          pnpm install
+          pnpm tauri build -- --target x86_64-apple-darwin
+          pnpm tauri build -- --target aarch64-apple-darwin
```
lucasfernog-crabnebula: note that this compiles separate apps for Intel and Apple Silicon; alternatively we could ship universal binaries too if that's desired (they are bigger because include the runtime for both architectures)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+          APPLE_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
+          APPLE_SIGNING_IDENTITY: ${{ secrets.APPLE_SIGNING_IDENTITY }}
+          APPLE_ID: ${{ secrets.APPLE_ID }}
+          APPLE_PASSWORD: ${{ secrets.APPLE_PASSWORD }}
+          APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
+          APPLE_API_ISSUER: ${{ secrets.APPLE_API_ISSUER }}
+          APPLE_API_KEY: ${{ secrets.APPLE_API_KEY }}
+          APPLE_API_KEY_PATH: api.p8
+
+          # using the same keys as v1
```
lucasfernog-crabnebula:
```suggestion
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+          TAURI_PRIVATE_SIGNING_KEY: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY }}
+          TAURI_KEY_SIGNING_PASSWORD: ${{ secrets.TAURI_KEY_SIGNING_PASSWORD }}
+
+      - name: upload assets
+        uses: crabnebula-dev/cloud-release@v0
+        with:
+          command: release upload ${{ env.CN_APPLICATION }} --framework tauri
+          api-key: ${{ secrets.CN_API_KEY }}
+          path: ./apps/desktop-solid/src-tauri
```
lucasfernog-crabnebula:
```suggestion
          working-directory: ./apps/desktop-solid/src-tauri
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,92 @@
...
+        run: cat apps/desktop-solid/.env
+
+      - name: Build MacOS Apps
+        # Install x86_64-apple-darwin for mac and build Tauri binaries
+        # Build both intel and apple silicon
+        working-directory: apps/desktop-solid
+        run: |
+          rustup target add x86_64-apple-darwin 
+          pnpm install
+          pnpm tauri build -- --target x86_64-apple-darwin
```
lucasfernog-crabnebula: don't really need the leading `--` here but ok

.github/workflows/publish.yml
atilafassina: I tried to be the least disruptive as possible. This step comes from the old app's CI.
@richiemcilroy

.github/workflows/publish.yml
atilafassina: same as the one above. This was already in the last version.

ping @richiemcilroy again

.github/workflows/publish.yml
atilafassina: ouch. that's a bad copy/paste from the docs I wrote 😅

good catch!

.github/workflows/publish.yml
richiemcilroy: I agree @lucasfernog-crabnebula - don't worry about it for now... appreciate @atilafassina working around my spaghetti code 🙈

.github/workflows/publish.yml
richiemcilroy: yeah just a debug step I think... this can be removed

Issue #91, opened 2024-10-03
Only require screen capture and accessibility permissions


richiemcilroy:
We need to make it so that only screen capture and accessibility permissions are required when starting the app for the first time.

Webcam/microphone should only be required when you go to select either of those options for the first time.

Feel free to DM me on the [Cap discord](https://cap.link/discord) to talk about this more.

PR #100, opened 2024-10-05
feat: only require screen capture and accessibility permissions

r17x:
## Changes

* Removed Audio (Microphone) Request permission on first-run.
* Request permission when user click Audio (Microphone) Selection - (only show when not `granted`).

### Attachments

| Granted | Not Granted |
|--------|--------|
| image | image |
|  image | |

/claim #91

Ready for review @richiemcilroy


richiemcilroy:
Hey @r17x this is awesome! Thanks for submitting so quickly. 

Would you be able to also do this for webcam as well?

Make it so that camera permission is not required at startup, until you decide to (by clicking the list). 

Also, willing to add another $100 to the bounty if you can get the green MacBook webcam light to stop showing when no webcam is currently selected.

r17x:
Already updated the PR @richiemcilroy 

Here's videos for another $100 bounty.

https://github.com/user-attachments/assets/28500477-9a33-43ea-bdc3-8ca8883d7d74

r17x:
Heres for camera (same behavior with `microphone`). @richiemcilroy 

https://github.com/user-attachments/assets/1dfdf7b8-599b-4915-bc5e-7aa11f0e041e

richiemcilroy:
Hey @r17x! looks awesome :)

One problem I've noticed is that once you do give permission, you only see the devices (webcam etc) once you right click and refresh.

Is there a way to make it so that we check again/refresh the list once the permission has been granted?

Also if it's possible to not open the dropdown list when we're actually requesting the permission.

So technically the entire "No Camera" button should request for permission on click, instead of just "Request Permission".

**Actions:**
	1.	Fix Device List Update: Automatically refresh the list of devices (webcam, etc.) once permission is granted, without requiring a manual right-click or refresh.
	2.	Prevent Dropdown from Opening During Permission Request: Ensure that the dropdown list for devices doesn’t open while the app is requesting camera permissions.
	3.	Update “No Camera” Button Behavior: Modify the “No Camera” button to request permission when clicked, instead of only the “Request Permission” button initiating the permission request.

apps/desktop/src/routes/(window-chrome)/index.tsx
```diff
@@ -292,12 +301,21 @@ export default function () {
...
-            commands.setRecordingOptions({
-              ...options.data,
-              cameraLabel: d.deviceId ? d.label : null,
-            });
+          onChange={(device: MediaDeviceInfo | null) => {
+            if (options.data?.captureTarget) {
+              if (device) {
+                commands.setRecordingOptions({
+                  ...options.data,
+                  cameraLabel: device && device?.deviceId ? device?.label : null,
```
galabra: `device && device?.deviceId` is equivalent to `device?.deviceId`

richiemcilroy:
Let me know your thoughts @r17x whenever you get a chance - would love to get this live ASAP

r17x:
@richiemcilroy Here's my new updates:

> 1. Fix Device List Update: Automatically refresh the list of devices (webcam, etc.) once permission is granted, without requiring a manual right-click or refresh.

Automatically update the list options devices (you must be click the options lists for showing the list of devices). Anyway, I didn't understood what's you mean `right-click` or `refresh`. 

> 2. Prevent Dropdown from Opening During Permission Request: Ensure that the dropdown list for devices doesn’t open while the app is requesting camera permissions.

Fixed

> 3. Update “No Camera” Button Behavior: Modify the “No Camera” button to request permission when clicked, instead of only the “Request Permission” button initiating the permission request.

Fixed


https://github.com/user-attachments/assets/fbac08c3-638f-4a08-ba9c-8e57eb7e9276

apps/desktop/src/utils/queries.ts
```diff
@@ -51,13 +61,42 @@ export function createWindowsQuery() {
 
 export function createAudioDevicesQuery() {
   const devices = createQuery(() => listAudioDevices);
-  createTimer(() => devices.refetch(), 1000, setInterval);
```
r17x, line 54: I removed this and set using `queryOptions -> refetchInterval: 1000` - same way

apps/desktop/src/routes/(window-chrome)/index.tsx
r17x: Nice catch, thank you @galabra

PR #104, opened 2024-10-06
fix: fix web url in README.md

c1sc0:
The url to the website was not working.


richiemcilroy:
Thank you! Previously we used to have a web recorder, but this no longer exists.

PR #106, opened 2024-10-08
fix(MediaStream): cleanup unused media stream

r17x:
## Changes

* removed unused stream. Every getUserMedia called it will be return new instance of stream and make light camera in real device will be turn on. in this pr just "mongkey patch" for cleanup unused stream. this pr isn't a better solution but I think could be fix the current issue for light on.

@richiemcilroy


Brendonovich (approved):
I meant to do this earlier, thankyou!

Issue #114, opened 2024-10-10
Add screen selector button


richiemcilroy:
In the main Cap window, add an option to select between different screens.

Like window selector, but for different screens.

5war00p:
@richiemcilroy I've raised a PR, please review and comment if anything else is needed.

webbdays:
seems merged.
bounty still open?

PR #118, opened 2024-10-11
feat: multiple screen selection

5war00p:
Contains rust and typescript file changes w.r.to Multi screen selection.

/claim https://github.com/CapSoftware/Cap/issues/114


richiemcilroy:
Hey @5war00p - tested this out and even though I can see the two different screens in the screen selector - no matter what one I pick, it always records my Macbook screen, not the monitor. Switched between extended mode / screen mirroring. Also, can you change the labels (e.g. Screen 1, Screen 2), to the actual monitor / screen names?

Also, if there's only 1 screen selected, don't allow the option to click select.

5war00p:
> Hey @5war00p - tested this out and even though I can see the two different screens in the screen selector - no matter what one I pick, it always records my Macbook screen, not the monitor. Switched between extended mode / screen mirroring. Also, can you change the labels (e.g. Screen 1, Screen 2), to the actual monitor / screen names?
> 
> Also, if there's only 1 screen selected, don't allow the option to click select.


I see me change those conditions for name, I don't think if we get names for screens, but will check that up.

Also the IDs of the screens, looks screen IDs not coming accurately.

Will debug it.

5war00p:
image

I've updated it

5war00p:
@richiemcilroy Did checked this again? I see there are lot of changes, and I resolved conflicts as well

PR #122, opened 2024-10-13
Optimize AsyncVideoDecoder for Better Memory (Possible memory leak fix)

newideas99:
# Pull Request: Optimize AsyncVideoDecoder for Better Memory Management and Performance

## Changes Made

1. Added new constants for cache management:
   ```rust
   const MAX_CACHE_MEMORY: usize = 1024 * 1024 * 1024; // 1 GB max cache size
   const CACHE_CLEANUP_INTERVAL: Duration = Duration::from_secs(60); // Clean up every 60 seconds
   ```

2. Introduced `CachedFrame` struct for better frame management:
   ```rust
   #[derive(Clone)]
   struct CachedFrame {
       frame_number: u32,
       frame: DecodedFrame,
       last_accessed: Instant,
   }
   ```

3. Changed cache data structure from `BTreeMap` to `VecDeque`:
   ```rust
   let mut cache = VecDeque::new();
   ```

4. Implemented new cache management methods:
   ```rust
   fn cleanup_cache(cache: &mut VecDeque<CachedFrame>, cache_size: &mut usize) {
       // ... (implementation)
   }

   fn aggressive_cleanup(cache: &mut VecDeque<CachedFrame>, cache_size: &mut usize) {
       // ... (implementation)
   }
   ```

5. Modified frame decoding and caching logic:
   - Introduced `frame_to_send` variable to store the frame to be sent.
   - Updated cache insertion and removal logic.
   - Implemented periodic cache cleanup.

6. Moved sender.send() outside of all loops:
   ```rust
   sender.send(frame_to_send).ok();
   ```

## Explanation

The main issues addressed in this update are related to memory management, performance optimization, and potential undefined behavior due to improper use of the oneshot sender.

### Prevention of Memory Leaks and Improved Performance

1. **Better Cache Management**: The new `CachedFrame` struct and `VecDeque` data structure allow for more efficient frame caching and access patterns. The `cleanup_cache` and `aggressive_cleanup` methods ensure that the cache doesn't grow beyond specified limits, preventing potential memory leaks.

2. **Single Use of Sender**: By moving the `sender.send()` call outside of all loops, we ensure that the sender is used exactly once per request. This prevents potential memory leaks that could occur if frames were sent multiple times or if the sender was dropped without being used.

3. **Efficient Frame Caching**: The `frame_to_send` variable allows us to store the required frame as soon as it's decoded, without immediately sending it. This approach prevents unnecessary decoding of frames after the target frame is found, reducing memory usage and improving performance.

4. **Periodic Cache Cleanup**: The implementation of periodic cache cleanup helps maintain a reasonable memory footprint over time, preventing the accumulation of unused frames in memory.

### Problems Addressed

1. **Memory Leaks**: The previous implementation risked accumulating too many frames in memory without a proper cleanup mechanism. The new cache management system addresses this.

2. **Inefficient Caching**: The old `BTreeMap` implementation was less efficient for the access patterns typically seen in video decoding. The new `VecDeque` structure is more appropriate for this use case.

3. **Potential Undefined Behavior**: The previous code had multiple points where the oneshot sender could be used, risking using it more than once or not at all in some code paths. The new implementation ensures consistent, single use of the sender.

4. **Performance Issues**: The old code continued decoding frames even after finding the requested frame. The new implementation breaks the decoding loop as soon as the required frame is found, improving performance.

## Conclusion

These changes significantly improve the robustness, efficiency, and safety of the `AsyncVideoDecoder`. By implementing better cache management, ensuring proper use of the oneshot sender, and optimizing the frame decoding process, we've mitigated the risk of memory leaks, improved overall performance, and enhanced the reliability of the video decoding system.

### DIFF
--- a/crates/rendering/src/decoder.rs
+++ b/crates/rendering/src/decoder.rs
@@ -1,9 +1,10 @@
 use std::{
     cell::Cell,
-    collections::BTreeMap,
+    collections::{BTreeMap, VecDeque},
     path::PathBuf,
     ptr::{null, null_mut},
     sync::{mpsc, Arc},
+    time::{Duration, Instant},
 };
 
 use ffmpeg::{
@@ -18,19 +19,27 @@ use ffmpeg_sys_next::{
     AVPixelFormat, AV_CODEC_HW_CONFIG_METHOD_HW_DEVICE_CTX,
 };
 
+const FRAME_CACHE_SIZE: usize = 50;
+const MAX_CACHE_MEMORY: usize = 1024 * 1024 * 1024; // 1 GB max cache size
+const CACHE_CLEANUP_INTERVAL: Duration = Duration::from_secs(60); // Clean up every 60 seconds
+
 pub type DecodedFrame = Arc<Vec<u8>>;
 
 enum VideoDecoderMessage {
-    GetFrame(u32, tokio::sync::oneshot::Sender<Option<Arc<Vec<u8>>>>),
+    GetFrame(u32, tokio::sync::oneshot::Sender<Option<DecodedFrame>>),
+}
+
+#[derive(Clone)]
+struct CachedFrame {
+    frame_number: u32,
+    frame: DecodedFrame,
+    last_accessed: Instant,
 }
 
 fn ts_to_frame(ts: i64, time_base: Rational, frame_rate: Rational) -> u32 {
-    // dbg!((ts, time_base, frame_rate));
     ((ts * time_base.numerator() as i64 * frame_rate.numerator() as i64)
         / (time_base.denominator() as i64 * frame_rate.denominator() as i64)) as u32
 }
-
-const FRAME_CACHE_SIZE: usize = 50;
 
 pub struct AsyncVideoDecoder;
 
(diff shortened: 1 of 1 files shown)


newideas99:
Everything seemed to build and work when i tested, but give it a test before pushing

Brendonovich:
Using `VecDeque` and switching to bounded channels are great changes, thanks. Wanted to note a couple of things:

> Single Use of Sender: By moving the sender.send() call outside of all loops, we ensure that the sender is used exactly once per request. This prevents potential memory leaks that could occur if frames were sent multiple times or if the sender was dropped without being used.

`sender` is a oneshot, it's impossible for more than one frame to get sent over it. Dropping the sender without using it is also fine, we treat it the same as sending `None`.

> Performance Issues: The old code continued decoding frames even after finding the requested frame. The new implementation breaks the decoding loop as soon as the required frame is found, improving performance.

This is done on purpose to warm the cache ahead of time during playback, we keep decoding past the requested frame until we reach the limit or until a new frame request comes in.

PR #159, opened 2024-11-08
Better audio support, step one

filleduchaos:
This primarily removes the dependency on the FFmpeg CLI for decoding audio + adds full support for stereo audio, but also introduces performance and quality improvements for audio rendering.

I'm still working on a good-enough muxer (which would let us drop the dependency on the CLI entirely), but I've decided to move that effort to its own branch for now since the work in this branch should be testable/releasable on its own.


Brendonovich (approved):
Haven't tested stereo but recording mono, playback, and exporting work great, nice work

PR #162, opened 2024-11-11
feat: Video importer (bring your own video)

richiemcilroy:
Create a new Cap project from any .mp4 video

mp4 only for now - since we'd probably have to build or hook up the encoder to support different formats?

https://github.com/user-attachments/assets/67709064-9eee-4ea5-bd6f-11d5461219e1


Brendonovich:
- drag and drop didn't work in my testing, it might need some tauri stuff to implement
- >30fps videos break the editor atm, either we should prevent higher fps videos or just make the editor work with arbitrary framerates before releasing this

PR #175, opened 2024-11-20
Add window-state plugin, Fix traffic light positioning…

ItsEeleeya:
#### Adds Window State plugin
With window-state, the app will remember the size and position of each window the next time they are opened (setup, window-capture-occluder, camera, prev-recordings are ignored).

* Use native decorations and shadow on in-progress-recording window again.
* Make editor titlebar smaller to match the rest of the windows
* Fixes traffic lights not being positioned correctly

#### Known Issues
* You might briefly see the traffic lights show in the incorrect location before the page is loaded.
* When windows are minimized by the app and you restore them, the theme set for the window won't be restored.

These issues will be fixed in a later PR with a more major refactor around the titlebars.


Brendonovich:
Works great 👍

PR #179, opened 2024-11-22
Fixes for Windows, cross-platform copy, system theme...

ItsEeleeya:
- [x] Use ~~arboard~~ [clipboard-rs](https://github.com/ChurchTao/clipboard-rs) for cross-platform clipboard writes.
(- Removes plugin-clipboard-manager)
- [x] Fix window bounds (Windows) to not include the window area's shadow
- [x] Minor fix to recording actors
- [x] Don't `unwrap` on `prev-recordings` window's `cursor_position()` as it might sometimes be an Err on windows.
- [x] `prev-recordings` on Chromium webviews have an issue where elements with a z-index don't respect the overflow style properly.
Previous:

 → 

<br>

### Other changes:
* Show export progress on taskbar/dock icon

* System theme support. Adds theme previews.
* Fixed an issue where the logo on the main window would jump in size when dark mode was on.
* Request user attention after an update has finished downloading.
 * Fix scrollbar colors not respecting app's theme. Use `thin` scrollbar
 
Fixed scrollbar style on Windows

 * New installer sidebar banner
MSI installer banner

 * New macos-specific icon matching the rest of the system
 

~~## Known Issue (Windows-only)~~
~~After logging into the app and restarting, trying to open the settings window will freeze the app, likely an issue with how the setting's store is read. For now during development you need to remove the old store and log back in.~~
_Magically_ Solved by @Brendonovich

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -2,9 +2,9 @@
   "name": "@cap/desktop",
   "type": "module",
   "scripts": {
-    "dev": "dotenv -e ../../.env -- pnpm run prepare && tauri dev",
-    "build:tauri": "dotenv -e ../../.env -- pnpm run prepare && tauri build",
-    "prepare": "node scripts/prepare.js",
+    "dev": "dotenv -e ../../.env -- pnpm run preparescript && tauri dev",
+    "build:tauri": "dotenv -e ../../.env -- pnpm run preparescript && tauri build",
+    "preparescript": "node scripts/prepare.js",
     "localdev": "dotenv -e ../../.env -- vinxi dev --port 3001",
     "build": "vinxi build",
     "tauri": "tauri"
--- a/apps/desktop/scripts/prepare.js
+++ b/apps/desktop/scripts/prepare.js
@@ -91,7 +91,6 @@ async function prepareFfmpegSidecar() {
       const progress = ((downloadedBytes / totalBytes) * 100).toFixed(2);
       process.stdout.write(`\rDownloading: ${progress}%`);
     }
-
     console.log("\nDownload complete.");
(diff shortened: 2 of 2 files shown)


Brendonovich:
Working well on macos for me, still need to test on windows

ItsEeleeya:
even after removing the previous store and logging back in, opening the settings freezes the entire app (including the frontend sometimes)
sreenshot of store.dat

ItsEeleeya:
@Brendonovich I updated the preparation script to work on windows. It now correctly adds the DLLs to the resources on Windows.
It defines a wix upgrade code as [recommended by Tauri](https://tauri.app/reference/config/#upgradecode).
If the version includes anything other than integers, it'll get transformed on windows.
 e.g: `0.3.3-beta.12` -> `0.3.3.12`
 stripDebugSymbols is macos only for now.
 
I had to update tauri-cli to `2.1.0` from `2.0.0-rc` and after that for some reason Turbo doesn't seem to be able to build properly for me. It suddenly needed a new config and even then it was failing.
**The pnpm-lock might be broken**

tauri build and some plugins are also on `rc`, maybe these should be updated as well?

ItsEeleeya:
Preparation scrip is now fixed, the github CI prebuild script now uses it for creating the platform configs (haven't tested it on Github).
Installing the app now correctly moves the required ffmpeg DLLs to its location.
This PR now sets a fixed upgrade code for the WIX installer as recommended by Tauri, it's the same GUID that tauri crates with the current `productName` from the configs which you can see with the command: `tauri inspect wix-upgrade-code`

Brendonovich (approved):

PR #217, opened 2024-12-18
Minor fixes for Windows

ItsEeleeya:
- [x] Fixes InProgressRecording window's drag area 
- [x] Temporarily fixes the window sizing and positioning on Windows
  - This fix depends on not using tauri-plugin-window-state on Windows for now.
 The root of this issue will be fixed in an upcoming release of Tauri with Tao 0.31.0 as shown here:
 https://github.com/tauri-apps/tao/pull/1017
 
- [ ] Selecting a virtual microphone freezes the app
- [ ] Recordings with a microphone don't stop


richiemcilroy:
Added some extra fixes, including mic support

Issue #224, opened 2024-12-19
Time/size estimation for export


richiemcilroy:
Add a time (how long it will take to export), and a size (how large the file will be), to below the export button.

This needs to be added to both the `prev-recordings` window thumbnails, and Editor.

Gyan-max:
### Brief Implementation Plan for Issue #224

1. **Locate Export Button in Code**  
   - Identify the components rendering the **export button** in:
     - **Prev-recordings window thumbnails**.
     - **Editor**.

2. **Add UI for Time & Size Display**  
   - Add placeholders below the export button:
     ```html
     <div class="export-info">
       <span class="export-time">Time: Calculating...</span>
       <span class="export-size">Size: Calculating...</span>
     </div>
     ```

3. **Implement Calculation Logic**  
   - Write a utility function to estimate:
     - **Export Time**: Based on video length and encoding speed.
     - **File Size**: Based on bitrate, resolution, and length.
     ```javascript
     function calculateExportInfo({ length, bitrate }) {
       const time = (length / 60) * 2; // Example: 2x speed
       const size = (length * bitrate) / 8 / 1024 / 1024; // MB
       return { time: `${Math.ceil(time)} min`, size: `${Math.ceil(size)} MB` };
     }
     ```

4. **Integrate Dynamic Updates**  
   - Hook the calculation logic into thumbnail and editor states:
     ```javascript
     const exportInfo = calculateExportInfo(videoProperties);
     setExportInfo(exportInfo);
     ```

5. **Styling**  
   - Ensure the display matches the existing design:
     ```css
     .export-info { font-size: 0.85em; color: gray; margin-top: 5px; }
     ```

6. **Testing**  
   - Test functionality in various scenarios:
     - Short/long videos.
     - Different resolutions and bitrates.

richiemcilroy:
Hey @onyedikachi-david. Thanks for this. Need to change it so that it does the following:

- The time estimation should be the longest of either `camera.mp4` or `display.mp4`. 
- Need to have an estimation of the rendered file size, not the current file size of the source file

let me know if you need anything else from me

richiemcilroy:
Hey @onyedikachi-david - did you see this?

onyedikachi-david:
> Hey @onyedikachi-david - did you see this?

I forgot, pushing some changes ASAP

PR #231, opened 2024-12-27
Add video metadata display

onyedikachi-david:
/claim #224
Fixes: #224

image


onyedikachi-david:
> Hey @onyedikachi-david. Thanks for this. Need to change it so that it does the following:
> 
> * The time estimation should be the longest of either camera.mp4 or display.mp4.
> * Need to have an estimation of the rendered file size, not the current file size of the source file
> 
> let me know if you need anything else from me


Hi, I have implemented them @richiemcilroy

onyedikachi-david:
Thanks for refining the export data estimates

PR #242, opened 2025-01-18
auth via deep link

### LINKED ISSUE #240: Urgent: Replace desktop auth with deep links
Currently, when signing in from desktop via web, you are redirected from the web app to a 127.0.0.x url.

This is supposed to communicate with the app via a local server, verifying the auth.

However, this method is not reliable and doesn't work all of the time, meaning the auth process does not work for some users.

Objective:
- replace current desktop auth process with more reliable method. 

One way to do this would be to redirect the user auth using deep links, attaching the token to the deep link url.

neo773:
This PR adds authentication via deep linking.

Deep link on macOS isn't supported at runtime so we still need to keep the port option for development purposes. It's toggled programmatically based on `VITE_ENVIRONMENT`

image

https://github.com/user-attachments/assets/55b3eaf4-3ce7-49c3-8928-a1b86411f1bb

/claim #240
/closes #240

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -244,7 +244,7 @@ dependencies = [
  "wayland-backend",
  "wayland-client",
  "wayland-protocols",
- "zbus",
+ "zbus 4.4.0",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


richiemcilroy:
Hey, looks great!

one thing though. When using locally, the auth callback which triggers the deep link opens the production version of Cap for me, meaning I can't sign in locally. For local dev, we should use the original method of auth. Deep links for production.

PR #347, opened 2025-02-27
Auto zoom implementation

tolotrasamuel:
https://github.com/user-attachments/assets/14295232-35ee-4ea9-9e29-783f1765fec2


Brendonovich:
This is a good start, but something I should have detailed in an issue is that the zoom shouldn't directly follow the cursor, rather it should lock onto a starting point, then move to another anchor point once the cursor moves close enough to the edge of the frame, with the motion being smoothed with the same physics simulation as the cursor.
I've not got around to implementing this myself as it's a bit tricky, but you're welcome to try it.

Issue #343, opened 2025-02-26
error when running pnpm dev:desktopi


tolotrasamuel:
### Description
 ```
 --- stderr
  ** BUILD FAILED **


  The following build commands failed:
        CompileC /Users/home/PhpstormProjects/Cap/target/debug/build/cidre-4d9ea21412eb6ba0/out/pomace.build/Debug/core_audio.build/Objects-normal/arm64/core_audio.o /Users/home/.cargo/git/checkouts/cidre-2c979cd9b38fdb3d/3479c18/cidre/pomace/core_audio/core_audio.m normal arm64 objective-c com.apple.compilers.llvm.clang.1_0.compiler (in target 'core_audio' from project 'pomace')
  (1 failure)

  thread 'main' panicked at /Users/home/.cargo/git/checkouts/cidre-2c979cd9b38fdb3d/3479c18/cidre/build.rs:145:5:
  assertion failed: status.success()
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
/Users/home/PhpstormProjects/Cap/apps/desktop:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @cap/desktop@ dev: `pnpm -w cap-setup && cross-env RUST_BACKTRACE=1 dotenv -e ../../.env -- pnpm run preparescript && tauri dev`
Exit status 101
 ELIFECYCLE  Command failed with exit code 101.
```
### Reproduction

I followed all steps in `### General Setup` from the `CONTRIBUTIONS.MD`

### Additional Context

- Cap version: Master branch

- Operating system, version: MACOS 14.0

**Comment by tolotrasamuel** [comment]
```
  /Users/samuel/.cargo/git/checkouts/cidre-2c979cd9b38fdb3d/3479c18/cidre/pomace/core_audio/core_audio.h:12:6: error: expected ';' after top level declarator
  Class CA_TAP_DESCRIPTION;
       ^
       ;
  /Users/samuel/.cargo/git/checkouts/cidre-2c979cd9b38fdb3d/3479c18/cidre/pomace/core_audio/core_audio.h:22:9: error: use of undeclared identifier 'CA_TAP_DESCRIPTION'
          CA_TAP_DESCRIPTION = NSClassFromString(@"CATapDescription");
          ^
[...]

Issue #345, opened 2025-02-26
error[E0502]: cannot borrow `*image_buf` as immutable because it is also borrowed as mutable


tolotrasamuel:
### Description

The source code is not building with the error: 
```

error[E0502]: cannot borrow `*image_buf` as immutable because it is also borrowed as mutable
  --> crates/rendering/src/decoder/avassetreader.rs:34:41
   |
32 |                     let _lock = image_buf.base_address_lock(LockFlags::READ_ONLY).unwrap();
   |                                 --------- mutable borrow occurs here
33 |
34 |                     let bytes_per_row = image_buf.plane_bytes_per_row(0);
   |                                         ^^^^^^^^^ immutable borrow occurs here
...
57 |                 } else {
   |                 - mutable borrow might be used here, when `_lock` is dropped and runs the `Drop` code for type `BaseAddrLockGuard`

error[E0502]: cannot borrow `*image_buf` as immutable because it is also borrowed as mutable
  --> crates/rendering/src/decoder/avassetreader.rs:35:33
   |
32 |                     let _lock = image_buf.base_address_lock(LockFlags::READ_ONLY).unwrap();
   |                                 --------- mutable borrow occurs here
...
35 |                     let width = image_buf.width() as usize;
   |                                 ^^^^^^^^^ immutable borrow occurs here
...
57 |                 } else {
   |                 - mutable borrow might be used here, when `_lock` is dropped and runs the `Drop` code for type `BaseAddrLockGuard`

error[E0502]: cannot borrow `*image_buf` as immutable because it is also borrowed as mutable
  --> crates/rendering/src/decoder/avassetreader.rs:36:34
   |
32 |                     let _lock = image_buf.base_address_lock(LockFlags::READ_ONLY).unwrap();
   |                                 --------- mutable borrow occurs here
...
36 |                     let height = image_buf.height();
   |                                  ^^^^^^^^^ immutable borrow occurs here
...
57 |                 } else {
   |                 - mutable borrow might be used here, when `_lock` is dropped and runs the `Drop` code for type `BaseAddrLockGuard`

error[E0502]: cannot borrow `*image_buf` as immutable because it is also borrowed as mutable
  --> crates/rendering/src/decoder/avassetreader.rs:40:29
   |
32 |                     let _lock = image_buf.base_address_lock(LockFlags::READ_ONLY).unwrap();
   |                                 --------- mutable borrow occurs here
...
40 |                             image_buf.plane_base_address(0),
[...]

PR #348, opened 2025-02-27
Added XCode Version Requirement in CONTRIBUTING.md

tolotrasamuel:
For MacOS, the desktop app will not build if the version is not `15.4 (Build version 15F31d)`.

This fixes https://github.com/CapSoftware/Cap/issues/343 and https://github.com/CapSoftware/Cap/issues/345


Brendonovich:
I have XCode 16.2, I think the important thing is that XCode is up to date for your system rather than being a specific version

Issue #312, opened 2025-02-12
Fix cursor recording timings


Brendonovich:
Cursor timings are slightly delayed compared to the screen capture, leading to innacurate rendering

PR #350, opened 2025-02-28
Fixes: Fix cursor recording timings

tolotrasamuel:
Fixes https://github.com/CapSoftware/Cap/issues/312


Brendonovich:
Not sure if this'll fully address the issue but it'll be a bit better

PR #448, opened 2025-04-20
Blur shader

oscartbeaumont:
This PR implements a wgpu blur shader. It does not handle the integration with the editor just implementing the filter in Rust.

You must uncomment the `BlurLayer::render` call to preview it.

An example of it bluring part of a frame (notice the PR title):

Screenshot 2025-04-20 at 4 07 16 pm


Brendonovich:
- `blur_radius` isn't being used
- Can it be done in one render pass?
- random noise to avoid un-blurring?

EChambeee:
What is the purpose of blurring part of the frame, and how can you do it?

PR #569, opened 2025-05-23
Enable multi-threaded rendering

richiemcilroy:
## Summary
- add `num_cpus` to rendering crate
- make `RenderSegment` clonable
- parallelise `render_video_to_channel` using a worker pool

## Testing
- `cargo check --workspace` *(fails: failed to load source for dependency `screencapturekit`)*


Brendonovich:
Will need to be reimplemented after the Skia rewrite

PR #571, opened 2025-05-25
Password protection for videos + Player UI tweaks

richiemcilroy:
## Summary
- Adds optional password protection to videos.
- New `password` column in `videos` table (encrypts and decrypts password)
- Hides metadata for video if no access
- Improved UI for the player, including better integrated chapters.


Brendonovich:
This PR doesn't yet password protect the actual API routes that give access to the video, we need to make sure the media files aren't accessible at all unless the password is provided

PR #587, opened 2025-06-02
FIX Resend "from" error for custom domain in ENV

guiguicdd:
# Problem
`serverEnv().RESEND_FROM_DOMAIN` is not correctly retrieving the value of `RESEND_FROM_DOMAIN`. Because of that, Resend returns a `422` error. This happens due to the piece of code that uses the domain URL as a fallback in `NEXT_PUBLIC_WEB_URL`.

https://github.com/CapSoftware/Cap/blob/b1094ab418069b46cace4bc0ae55c16746f72bfc/packages/database/emails/config.ts#L43

## Reproducing the Issue
[Railway](https://railway.com/deploy/PwpGcf) .ENV configuration
image

### Example of a request made by Cap web
```JSON
{
    "to": "someone@example.com",
    "from": "auth@https://cap-web-production.up.railway.app",
    "html": "<BODY>",
    "subject": "Your Cap Login Link"
}
```
### Response from Resend
```JSON
{
    "name": "validation_error",
    "message": "Invalid `from` field. The email address contains a colon.",
    "statusCode": 422
}
```

### Expected request to be made from Cap web with this fix
```JSON
{
    "to": "someone@example.com",
    "from": "auth@resend.example.com",
    "html": "<BODY>",
    "subject": "Your Cap Login Link"
}
```

# Possible Solution
Implement a more robust check and reference the variable globally, as the .ENV for RESEND_API_KEY is working correctly this way. If none of the options are available, an error should be thrown and the request to Resend should not be made.

I wasn’t able to test this change, so someone with a properly configured environment will need to test this solution or propose a better one for this issue.

If you're looking for this fix as well, let us know by reacting with 👀
If you have tested this solution and it worked, please react with 👍
otherwise, feel free to engage by proposing a solution

### DIFF
--- a/packages/database/emails/config.ts
+++ b/packages/database/emails/config.ts
@@ -5,6 +5,8 @@ import { Resend } from "resend";
 export const resend = () =>
   serverEnv().RESEND_API_KEY ? new Resend(serverEnv().RESEND_API_KEY) : null;
 
+const resendCustomDomain = serverEnv().RESEND_FROM_DOMAIN ? serverEnv().RESEND_FROM_DOMAIN : buildEnv.NEXT_PUBLIC_WEB_URL
+
 // Augment the CreateEmailOptions type to include scheduledAt
 type EmailOptions = {
   from: string;
@@ -39,9 +41,7 @@ export const sendEmail = async ({
       ? "Richie from Cap <richie@send.cap.so>"
       : buildEnv.NEXT_PUBLIC_IS_CAP
       ? "Cap Auth <no-reply@auth.cap.so>"
-      : `auth@${
-          serverEnv().RESEND_FROM_DOMAIN ?? buildEnv.NEXT_PUBLIC_WEB_URL
-        }`,
+      : `auth@${resendCustomDomain}`,
     to: test ? "delivered@resend.dev" : email,
     subject,
     react,
--- a/packages/database/emails/config.ts
+++ b/packages/database/emails/config.ts
@@ -5,7 +5,7 @@ import { Resend } from "resend";
 export const resend = () =>
   serverEnv().RESEND_API_KEY ? new Resend(serverEnv().RESEND_API_KEY) : null;
 
-const resendCustomDomain = serverEnv().RESEND_FROM_DOMAIN ? serverEnv().RESEND_FROM_DOMAIN : buildEnv.NEXT_PUBLIC_WEB_URL
+const resendCustomDomain = serverEnv().RESEND_FROM_DOMAIN ? serverEnv().RESEND_FROM_DOMAIN : null
 
 // Augment the CreateEmailOptions type to include scheduledAt
 type EmailOptions = {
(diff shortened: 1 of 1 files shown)


Brendonovich:
If `RESEND_FROM_DOMAIN` is already set in Railway then why would it be falling back to `WEB_URL` in the first place?
Have you actually experienced this error? [Resend's docs](https://resend.com/docs/api-reference/errors) say that the `validation_error` error uses status code 400 or 403, which doesn't match with that you've provided.

guiguicdd:
Their documentation seems inconsistent or incomplete. Since it's open source, I was able to find the piece of code responsible for returning a `422` with `validation_error`.

https://github.com/resend/resend-python/blob/3ba6b384db035b193bdbfdb8dd79d6b3badd08a4/resend/exceptions.py#L168

Yes, I did experience this strange bug with the current code. And what’s even more odd is that `RESEND_API_KEY` worked fine.

What I noticed is that `RESEND_API_KEY` was defined in a global scope, while `RESEND_FROM_DOMAIN` was only defined at the time of the function call. So I believe this adjustment should fix the issue.

I wasn’t able to reproduce the problem since it's not my main focus. I submitted the fix mainly to open up a discussion about how to properly solve this issue. Even if my current solution isn't 100% correct.

If you can test this fix in your test environment, or even try reproducing the problem by creating an account on Railway and configuring a custom domain (where the custom domain in Resend doesn't match the one set in the Railway web app), then try submitting a login form with an email address.

This error should appear in your Resend logs:

*(I edited the "to" and "from" fields using DevTools to match the example above in terms of the Railway app ID and my custom domain)*  
image

Following these steps, you should be able to reproduce the exact same error I experienced.

Thanks for checking the accuracy of the information.  
Yes, it does seem very much like a human error. I’ve looked deeply into everything on my end and couldn’t find another solution so far.
Except for the one I proposed here. But the issue is definitely happening in this specific scenario

Brendonovich:
I've not been able to reproduce this problem - have you redeployed the Cap service since updating the variables? The purple colour indicates they haven't been applied yet.

guiguicdd:
Yes. redeployed it with the needed changes. It's just for demonstration purposes. I'm not allowed to share real data. Just an example of the real problem. So I edited to better flow the examples I've used before on this PR.

Brendonovich:
I'm pretty sure this isn't a problem with Cap - I've got a new version of the Docker image sending emails via Resend just fine with no change to the email code.
image

image

You can test yourself with the `ghcr.io/capsoftware/cap-web:beta` image

guiguicdd:
Thank you very much. It worked here with this version of the Docker image, so I believe there might still be an issue.

The Docker image I was using was `ghcr.io/capsoftware/cap-web:latest`, while the one you sent me that actually worked is `ghcr.io/capsoftware/cap-web:beta`. I had to implement the changes manually because the migrations didn’t work when I simply updated the image version in the [existing deploy Railway template](https://railway.com/deploy/PwpGcf).

Either way, thank you!

It might be worth updating the [Railway template](https://railway.com/deploy/PwpGcf) to use this beta version of cap-web.

Brendonovich:
A new version will be made available at `latest` soon, I've just published a `beta` version that successfully migrates from `latest` - i'm not sure how you'll go migrating your current db since the changes are incompatible with the previous beta. You might need to adjust the migrations table to only have the first migration, and remove some of the new tables + rename some columns from `space*` to back to `organization*` for the new beta to migrate properly.
I'm gonna close this PR since Resend seems to be working now.

PR #712, opened 2025-07-02
Fix: Correctly generate thumbnail URLs for path-style and virtual-hosted S3 providers

lukmay:
This PR resolves an issue where image thumbnails fail to load when using a self-hosted, S3-compatible object storage provider with a public custom domain. The fix updates the thumbnail generation logic to correctly handle both path-style and virtual-hosted (subdomain) style URLs based on environment variables.

#### The Problem

When self-hosting with an S3-compatible provider that uses a public custom domain (e.g., Cloudflare R2), the `/api/thumbnail` endpoint was not correctly using the public URL from the `CAP_AWS_BUCKET_URL` environment variable. Instead, it would fall back to generating pre-signed URLs or incorrect raw URLs, causing the upstream image fetch by the Next.js server to fail with a `404 Not Found` error.

This resulted in broken image thumbnails across the dashboard:

image

#### The Solution

The `/api/thumbnail/route.ts` file has been updated to implement more robust URL generation logic:

*   The API route now uses the `S3_PATH_STYLE` environment variable as a switch to determine the correct URL structure.
*   If `S3_PATH_STYLE` is `true` (for providers like MinIO), it constructs a path-style URL: `http://<endpoint>/<bucket>/<key>`.
*   If `S3_PATH_STYLE` is `false` (for providers like Cloudflare R2), it constructs a virtual-hosted style URL using the public `CAP_AWS_BUCKET_URL`: `https://<custom_domain_url>/<key>`.
*   This ensures the correct public URL is always used when a custom storage configuration is provided, making the application compatible with a wider range of self-hosting setups.

#### How to Test

1.  Configure a self-hosted environment using an S3-compatible provider and a public custom domain.
2.  Set the `S3_PATH_STYLE` and `CAP_AWS_BUCKET_URL` environment variables according to your provider's requirements.
3.  Build the Docker image with this code change.
4.  Launch the application and upload a new video.
5.  Verify that the video thumbnail loads correctly and that there are no `404` errors related to image fetching.

This change significantly improves the self-hosting experience for a wider range of S3-compatible storage providers. Thank you for considering this contribution.

### DIFF
--- a/apps/web/app/api/thumbnail/route.ts
+++ b/apps/web/app/api/thumbnail/route.ts
@@ -41,7 +41,7 @@ export async function GET(request: NextRequest) {
     return new Response(
       JSON.stringify({ error: true, message: "Video does not exist" }),
       {
-        status: 401,
+        status: 404,
         headers: getHeaders(origin),
       }
     );
@@ -52,7 +52,7 @@ export async function GET(request: NextRequest) {
     return new Response(
       JSON.stringify({ error: true, message: "Video not found" }),
       {
-        status: 401,
+        status: 404,
         headers: getHeaders(origin),
       }
     );
--- a/apps/web/app/api/thumbnail/route.ts
+++ b/apps/web/app/api/thumbnail/route.ts
@@ -60,43 +60,45 @@ export async function GET(request: NextRequest) {
 
   const { video } = result;
   const prefix = `${userId}/${videoId}/`;
-
+  const screenshotPath = `${prefix}screenshot/screen-capture.jpg`;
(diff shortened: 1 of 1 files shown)


Brendonovich:
Cap Web supports specifying `S3_INTERNAL_ENDPOINT` and `S3_PUBLIC_ENDPOINT` so that `getSignedObjectUrl` works properly, which will respect `S3_PATH_STYLE`. Would you mind trying if it works for your case? We'd like to keep all S3 operations going through `bucketProvider` as much as possible.

image

PR #734, opened 2025-07-08
fix: mobile menu visible on large screens

abhix4:
this pr fix :
- the mobile menu which is intended to show only on small screens is visible on large screens also.
image


ameer2468:
Thank you for the PR!

PR #777, opened 2025-07-16
Native camera preview

oscartbeaumont:
Currently we use a websocket to send the camera preview to the webview and render it. Users have noticed sometimes it can lag and the existing system is also not as efficient. This new system renders the camera preview directly to the window using wgpu so it can stay fully in Rust and take full advantage of the GPU.

This PR builds on top of #772 

 - [x] Render wgpu within Tauri
 - [x] Cleanup Tauri state handling
 - [x] Get the frame data into the render event loop
 - [x] Pipe camera frame to wgpu
 - [x] Camera shader so it renders in the correct shape
 - [x] Position output correctly on screen
 - [x] Webview toolbar should remain. Currently it's gone.
 - [x] Tauri drag region across whole thing
 - [x] Resampler output pixel format (we gonna lock this to RGB for now)
 - [x] Move toolbar state to Rust (including keep it persistant)
 - [x] Properly size and position the window
 - [x] Offset camera from the top to account for the top bar
 - [x] Toolbar buttons workings
 - [x] Shader should render proper shape based on configuration
 - [x] Shader should render mirroed
 - [x] ffmpeg crashing when using toolbar buttons
 - [x] Camera output resolution is bad (+ ghosting)
 - [x] Can't reopen the camera preview once it's closed
 - [x] Error handling
 - [x] Pink backbackground/invalid position on load is wrong
 - [x] Cleanup
    - [x] When no image is present we should render transparent not pink
    - [x] Cleanup useless log messages + comments (thanks Cursor :()
    - [x] Put shader in a dedicated file
    - [x] Merge `camera.rs` and camera_preview.rs`
 - [x] Remove all websocket system
 - [x] `CameraPreview` should go on `App` (not possible but did the closest best thing)
 - [x] Implement `unmanage` properly
 - [x] Bring back camera loading screen
 - [x] Output is laggy
    - [x] Reuse the resampler output frame
    - [x] Reuse the wgpu texture
    - [x] Can we reduce `Mutex` contention?

This PR makes use of ffmpeg rescaler. We are likely going to move off this in the future but for now it makes shipping this change much easier.

### DIFF
--- a/apps/desktop/src-tauri/src/camera.rs
+++ b/apps/desktop/src-tauri/src/camera.rs
@@ -1,15 +1,16 @@
 use anyhow::Context;
-use cap_media::{feeds::RawCameraFrame, frame_ws::WSFrame};
+use cap_media::feeds::RawCameraFrame;
 use ffmpeg::{
     format::{self, Pixel},
     frame,
     software::scaling,
 };
-use flume::{Receiver, Sender};
-use futures::executor::block_on;
+use flume::Receiver;
+use futures::{executor::block_on, future::Either};
 use serde::{Deserialize, Serialize};
 use specta::Type;
 use std::{
+    pin::pin,
     sync::{
         Arc,
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


richiemcilroy:
niceee

apps/desktop/src-tauri/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "cap-desktop"
 version = "0.3.60"
 description = "Beautiful screen recordings, owned by you."
 authors = ["you"]
-edition = "2021"
+edition = "2024"
```
oscartbeaumont, line 6: This was done to allow the usage of if-let chains for more readable code.

This does require bumping the MSRV to 1.88.

PR #783, opened 2025-07-19
feat: add functionality to update pretty name in editor

diwash007:
Users can now rename their recordings right from the editor so that they can have their own identifier for indentifying recordings.

I have no previous background in Tauri but I went ahead and implemented this.
Please do let me know if this is the correct way to architect this feature or I should follow any other specific paradigm?


Brendonovich:
Made a couple of changes but overall this was well done, thankyou!

Issue #707, opened 2025-07-02
Self-hosting Failed to upload recording


lonagi:
### Description

I have working cap-web and minio storage with external mariadb. in db there are:
auth api keys 1 , organisation members 1, users 1, drizzle migrations 2 rows, another tables have 0.
Authorised in my self-hosted cap and desktop mac app. Changed url in settings and after record video and upload it says failed. 

![Image](https://github.com/user-attachments/assets/2478e57b-271e-4d72-b50d-8778ea7ad106)

### Reproduction

Authorise in desktop app. Apply self-hosted settings. Record video and upload.

### Additional Context

- Cap version: latest
- Operating system, version: OS: Debian GNU/Linux 11 (bullseye) x86_64 

my dc file:
```yml
services:
  cap-web:
    container_name: cap-web
    image: ghcr.io/capsoftware/cap-web:latest
    restart: unless-stopped
    environment:
      DATABASE_URL: 'mysql://planetscale:test@mariadb:3306/planetscale?ssl=false'
      WEB_URL: https://mysite.com
      NEXTAUTH_URL: https://mysite.com
      # CHANGE THESE TO YOUR OWN VALUES
      DATABASE_ENCRYPTION_KEY: c7a1e98e1e5e4cec5a3fbf5eaf502d262fb99807ae2be8ee70537409e29cb6f9
      NEXTAUTH_SECRET: c7a1e98e1e5e4cec5a3fbf5eaf502d262fb99807ae2be8ee70537409e29cb6f9
      # AWS SDK V3 configuration for MinIO
      AWS_ACCESS_KEY_ID: capS3root
      AWS_SECRET_ACCESS_KEY: capS3root
      AWS_DEFAULT_REGION: us-east-1
      AWS_REGION: us-east-1
      AWS_ENDPOINT_URL_S3: http://minio:3902
      AWS_S3_USE_PATH_STYLE_ENDPOINT: "true"
      AWS_S3_FORCE_PATH_STYLE: "true"
      
      # Legacy AWS SDK V2 configuration
      AWS_ENDPOINT_URL: http://minio:3902
      AWS_S3_ENDPOINT: http://minio:3902
      
      # Cap-specific variables
      CAP_AWS_ACCESS_KEY: capS3root
      CAP_AWS_SECRET_KEY: capS3root
      CAP_AWS_BUCKET: pgbmk368j5v8rpw
      CAP_AWS_REGION: us-east-1
[...]

Issue #706, opened 2025-07-01
wep app cannot run MIGRATION_FAILED, HANDSHAKE_NO_SSL_SUPPORT


lonagi:
### Description

I just got your template dc file and edited a bit. Also I use external mariadb, and dont run ps-mysql service. Also I have dockerised nginx-proxy just using it to deploy different many site on my dedicated server. run dc up -d and getting errors app isn't starting

### Reproduction

docker compose:
```yml
services:
  cap-web:
    container_name: cap-web
    image: ghcr.io/capsoftware/cap-web:latest
    restart: unless-stopped
    environment:
      DATABASE_URL: 'mysql://planetscale:testpassword@mariadb:3306/planetscale?ssl={"rejectUnauthorized":false}'
      WEB_URL: http://localhost:3000
      NEXTAUTH_URL: http://localhost:3000
      # CHANGE THESE TO YOUR OWN VALUES
      DATABASE_ENCRYPTION_KEY: c7a1e98e1e5e4cec5a3fbf5eaf502d262fb99807ae2be8ee70537409e29cb6f9
      NEXTAUTH_SECRET: c7a1e98e1e5e4cec5a3fbf5eaf502d262fb99807ae2be8ee70537409e29cb6f9
      # CHANGE THESE TO MATCH YOUR MINIO or S3 STORAGE SETUP
      CAP_AWS_ACCESS_KEY: capS3root
      CAP_AWS_SECRET_KEY: capS3root
      CAP_AWS_BUCKET: capso
      CAP_AWS_REGION: us-east-1
      S3_PUBLIC_ENDPOINT: http://localhost:3902
      S3_INTERNAL_ENDPOINT: http://minio:3902

      # nginx vhost
      VIRTUAL_HOST: domain.site.com
      LETSENCRYPT_HOST: domain.site.com
      LETSENCRYPT_EMAIL: email.com
    networks:
      - nginx-proxy
    # ports:
      # - 3000:3000

  # Local S3 Strorage
  minio:
    container_name: cap-minio-storage
    image: "bitnami/minio:latest"
    restart: unless-stopped
    # ports:
      # - "3902:3902"
      # - "3903:3903"
    environment:
      - MINIO_API_PORT_NUMBER=3902
      - MINIO_CONSOLE_PORT_NUMBER=3903
      # CHANGE THESE TO YOUR OWN VALUES
      - MINIO_ROOT_USER=capS3root
      - MINIO_ROOT_PASSWORD=capS3root
[...]

PR #805, opened 2025-07-25
fix: video playback failure

mogita:
Changes:

- [x] Parse `data.source` when it's a JSON string instead of an object, causing incorrect `videoSrc` returned, and eventually causing the video playback to fail when the type if `desktopMP4` (the browser would be guided to attempt the m3u8 file, while in case of desktopMP4 no m3u8 was generated).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,6 +50,7 @@ NEXT_PUBLIC_CAP_AWS_BUCKET=capso
 NEXT_PUBLIC_CAP_AWS_REGION=us-east-1
 # If you're not using AWS S3, point this to your custom S3 provider (default points to MinIO created by `pnpm dev`)
 NEXT_PUBLIC_CAP_AWS_ENDPOINT=http://localhost:3902
+CAP_AWS_ENDPOINT=${NEXT_PUBLIC_CAP_AWS_ENDPOINT}
 # If using a proxy such as CloudFront, set this the full URL of the bucket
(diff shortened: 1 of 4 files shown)


ameer2468:
Hey Mogita, thank you for your contributions and this one too!

Looks good to me.

Brendonovich:
In what cases could Drizzle provide `source` as a string rather than a JS object? Seems like a bug in Drizzle if that's happening

mogita:
> In what cases could Drizzle provide `source` as a string rather than a JS object? Seems like a bug in Drizzle if that's happening

I was wondering the same thing and found an overlook on my side. I'm running the dashboard with MariaDB which doesn't have native JSON type and stores JSON data in `LONGTEXT` type. Drizzle doesn't provide automatic conversion from `LONGTEXT` to an object and would expect the application to do a JSON.parse().

It seems that MariaDB is not officially supported by the project except that it was brought up a few times by the community #707 #706. For the next step:

- I created a custom JSON type for Drizzle to handle both MySQL's JSON and MariaDB's LONGTEXT, tested and it worked well. I can make another PR;
- or let me revert this PR and switch to MySQL only

Which approach would you go with?

Brendonovich:
Seeing as Drizzle doesn't have reliable support for MariaDB I think we'll stick with MySQL-only for now. If there's enough demand for it then we could reconsider in future - none of us are familiar with MariaDB atm.

mogita:
Certainly. I'll revert the app code and add a few words for the DB support in readme.

PR #850, opened 2025-08-06
fix(camera): ensure RGBA order by swapping R and B channels before se…

osamarehman:
The backend fix for the camera RGB overlay/distortion bug

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the color channel order in camera frame data to ensure accurate color representation when viewing images.


Brendonovich:
This may fix some cameras but it almost definitely will break others that are already working.
The next release will use our entirely new camera backend (the `camera-*` crates) which has much better pixel format handling, so this issue shouldn't be as common.

PR #858, opened 2025-08-07
Cleanup Rust warnings + edition 2024 & enable lints on all crates

oscartbeaumont:
We have soooooooooo many warnings it's sometimes hard to find compiler errors as my terminal can literally truncate them. It also makes Zed's problem view useless for me.

Clippy was also able to find a couple of bugs. Having all these warnings fixed means we can actually use it to spot bugs in CI instead of the bugs getting lost in the noise. Feels like a perfect fit given the current focus is stablity.

I also bumped all crates to the Rust 2024 edition. A couple are already 2024 so this won't affect the MSRV.

### DIFF
--- a/apps/desktop/src-tauri/src/audio_meter.rs
+++ b/apps/desktop/src-tauri/src/audio_meter.rs
@@ -82,9 +82,10 @@ struct MinNonNan(f64);
 
 impl Eq for MinNonNan {}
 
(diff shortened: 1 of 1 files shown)


oscartbeaumont:
@coderabbitai pause

It's going to be painful if we have it review *the entire codebase*

Brendonovich:
lgtm
image

PR #874, opened 2025-08-12
New recording flow improvements

oscartbeaumont:
- [x] Enable new recording flow in dev by default
 - [x] Fix issue with screen information being the hovered screen instead of the current screen
 - [x] Fix corrdinate logic for window hover one external monitors
 - [x] Display names are still not friendly enough on Windows
 - [x] Reduce `windows` features on `cap-displays`
 - [ ] ~~Is the Windows `z-index` implementation good enough?~~
 - [x] Implement the new UI flow
 - [x] Area select is showing up on each monitor. It should only be on a single one.
 - [x] When I open up the app, it auto selects the last option? e.g. if I used the area picker last time, if I reopen the app, it'll open with the area view.
 - [ ] Fix area selection not actually capturing on external monitors
 - [ ] macOS monitor and window bounds calculation are wrong. This was introduced when I did the Windows multimonitor work.
 - [ ] Windows hovering on application window seems to be buggy
      - [ ] It's detecting Webview 2 a lot which I think is our overlay so we might need to review that logic.

## Summary by CodeRabbit

- New Features
  - Target selection overlay now displays detailed screen information, including name, physical size, and refresh rate when selecting a display.
  - Settings > Experimental: “New recording flow” toggle is now visible for all users, allowing easier discovery and control of the feature.
  - Enhanced display info retrieval enables richer, more accurate details in the overlay during screen selection.


apps/desktop/src-tauri/src/target_select_overlay.rs
```diff
@@ -68,20 +67,42 @@ pub async fn open_target_select_overlays(
...
+                        let display_relative_bounds = if let Some(current_display) = &display {
+                            let display_position = current_display.raw_handle().logical_position();
+                            let relative_position = LogicalPosition::new(
+                                bounds.position().x() - display_position.x(),
+                                bounds.position().y() - display_position.y(),
+                            );
+                            let converted_bounds = LogicalBounds::new(relative_position, bounds.size());
+
+                            // Debug logging for multi-monitor coordinate conversion
+                            tracing::debug!(
```
oscartbeaumont: Remove this if the logic is good

PR #875, opened 2025-08-12
Refactor to use React Query mutations

Brendonovich:
The commit replaces local loading states and manual error handling with React Query mutations across multiple components for better state management and consistency.


apps/web/app/(org)/dashboard/caps/components/CapCard/CapCard.tsx
```diff
@@ -97,31 +98,50 @@ export const CapCard = ({
...
+    },
+    onSettled: () => {
       setConfirmOpen(false);
-    }
-  };
+    },
+  });
+
+  const duplicateMutation = useEffectMutation({
+    mutationFn: () => withRpc((r) => r.VideoDuplicate(cap.id)),
```
oscartbeaumont: I think this needs a run `EffectRuntime.runPromise`

PR #885, opened 2025-08-13
Use native camera preview on supported platforms

oscartbeaumont:


oscartbeaumont:
This is kinda sketchy. Let's just get it properly working with the new DComp swapchain stuff.

PR #1033, opened 2025-09-19
Upgrade Next.js + enable Turbo

oscartbeaumont:
Me don't like waiting for Next.js to compile. Upgrade to the latest version so we can potentially enable Turbopack?


oscartbeaumont:
@coderabbitai don't review this pr

PR #1062, opened 2025-09-24
fall back to software H264 encoder when hardware ProcessInput fails

richiemcilroy:
Windows capture now watches for the ProcessInput E_FAIL from hardware H.264, immediately swaps to a software-only encoder, and replays the stalled frame. That automatic fallback relies on a tiny helper tweak that lets us enumerate software MFTs on demand, so recording stays alive even on fragile GPUs.

- Detect the Media Foundation ProcessInput E_FAIL (0x80004005) that some Windows GPUs hit mid-stream and rebuild the encoder in software mode, replaying the stalled frame so recording keeps going
  - Allow the encoder helper to enumerate MFTs with caller-specified flags so we can force software transforms during the fallback
  - Share capture config up front so both studio and instant pipelines can reuse it when re-initialising the encoder

## Summary by CodeRabbit

- New Features
  - Scaled output support with distinct input/output resolutions.
  - New software-only encoder options and convenience constructors.

- Improvements
  - More flexible encoder/device selection with configurable enumeration flags and smarter filtering.
  - Dynamic hardware/software transform control and improved selection behavior.
  - Enhanced frame-timing and event-loop handling across Windows capture modes.

- Bug Fixes
  - Automatic software fallback on encoder failures to avoid dropped frames.
  - Fewer initialization issues across varied hardware.


richiemcilroy:
@coderabbitai review

PR #1092, opened 2025-09-28
Local upload tracking

oscartbeaumont:
TODO:
 - [ ] `do_presigned_upload` typed error
 - [ ] `do_presigned_upload` reporting progress to frontend
 - [ ] `do_presigned_upload` reporting progress to cloud
 - [ ] Cleanup imports in `upload.rs`
 - [ ] Can we store failed uploads on the project itself so we have it when the app restarts???
 - [ ] `upload_image` needs progress reporting
 - [ ] Is `progressive_upload` optional for instant mode???
 - [ ] Retry and backoff on the `do_presigned_upload` function. I think the multipart upload already has this logic so can we share it?
 - [ ] Progress being reported to the frontend

### DIFF
--- a/apps/desktop/src-tauri/src/recording.rs
+++ b/apps/desktop/src-tauri/src/recording.rs
@@ -30,10 +30,7 @@ use crate::{
     general_settings::{GeneralSettingsStore, PostDeletionBehaviour, PostStudioRecordingBehaviour},
     open_external_link,
     presets::PresetsStore,
-    upload::{
-        InstantMultipartUpload, build_video_meta, create_or_get_video, prepare_screenshot_upload,
-        upload_video,
-    },
+    upload::{InstantMultipartUpload, build_video_meta, create_or_get_video, upload_video},
     web_api::ManagerExt,
     windows::{CapWindowId, ShowCapWindow},
 };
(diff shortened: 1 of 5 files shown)


apps/desktop/src-tauri/src/recording.rs
```diff
@@ -840,21 +861,9 @@ async fn handle_recording_finish(
...
-
-    let meta = RecordingMeta {
-        platform: Some(Platform::default()),
-        project_path: recording_dir.clone(),
-        sharing,
-        pretty_name: format!("{target_name} {date_time}"),
-        inner: meta_inner,
-    };
-
+    // TODO: Can we avoid reloading it from disk by parsing as arg?
```
oscartbeaumont: Do this if we are confident in this approach.

oscartbeaumont:
Backup of random code I may want later.

```rust
#[derive(Default, Debug)]
pub enum UploadPartProgress {
    #[default]
    Presigning,
    Uploading {
        uploaded: i64,
        total: i64,
    },
    Done,
    Error(String),
}

#[derive(Default, Debug)]
pub struct UploadVideoProgress {
    video: UploadPartProgress,
    thumbnail: UploadPartProgress,
}
```

oscartbeaumont:
Follow-up PRs:
 - Unfeature flag upload progress in web on a PR ontop of the existing stuff for testing
 - Bulk create presigned URL's in chunks (due to expiry on them). Will greatly improve upload speed and Vercel function invocations.
 - Using multipart upload for studio mode
 - Allow uploading multiple chunks at the same time
 - Allow opening corrupted studio recordings with more than one part. Warn the user before opening.
 - Resume multipart uploads on startup instead of just restarting
 - Upload status on the thumbnail cause it's done out of band of the main uploader? -> I don't know if this is worth solving???

oscartbeaumont:
Known issues:
 - [x] Video player controls missing after upload
 - [x] Don't skip the first chunk, reupload it! We will need to dedup the parts array.
 - [x] Instant mode seems to not be yielded chunks for a long time (until ended maybe). This could be the new logic on chunk size?
 - [x] Windows issues (might be a problem from `main`)
 - [x] Retry policy on S3 request isn't working well. I've changed the config but it needs testing.

PR #1066, opened 2025-09-25
Populate `orgId` col on `videos` table

oscartbeaumont:
For now we are just gonna start assigning it for new uploads. All existing uploads will be to be backfilled separately.

Actions:
 - [x] Add organisation setting for default orgId
 - [x] Hook the default up to `/api/desktop/videos/create`

Future backfill PR:
 - [ ] Fill `orgId` on all videos
 - [ ] Make it required

https://app.planetscale.com/cap/cap-production/deploy-requests/38

## Summary by CodeRabbit

* **New Features**
  * Organization-scoped video uploads and org-aware desktop API; default-organization selector in Account Settings with save/refresh and unsaved-change guard.
  * Redesigned Select component with improved composition and styling.

* **Bug Fixes**
  * Clear warning and abort when uploading without an active organization.
  * Video creation and upload flows validate and apply organization context, updating user defaults when needed.

* **Chores**
  * Added defaultOrgId to users, DB config tweak, orgId backfill utility and workspace entry, and minor global CSS cleanup.

PR #1105, opened 2025-10-01
Enforce nullable constraints

oscartbeaumont:
Our previous definitions of `nanoId` and `encryptedText` in our database schema use Typescript to mark them as being required fields, but in practice that constrain isn't being enforced in the actual SQL DB schema. This means the types are lying. I was about to check and we have a non-insignificant amount of records which break these constrains which would be resulting in runtime errors for users trying to use Cap.

This PR updates the Drizzle schema to enforce these constrains and also relaxes the Typescript type on some fields to align with the reliability of the data in the database.

Blockers:
 - [x] We need #1066 merged and `orgId` to be required on `videos`
 - [ ] Figure out notifications

### DIFF
--- a/apps/web/lib/Notification.ts
+++ b/apps/web/lib/Notification.ts
@@ -88,7 +88,6 @@ export async function createNotification(
 			const [recipientUser] = await db()
 				.select({
 					preferences: users.preferences,
-					activeOrganizationId: users.activeOrganizationId,
 				})
 				.from(users)
 				.where(eq(users.id, recipientId))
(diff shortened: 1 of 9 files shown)


apps/web/lib/Notification.ts
```diff
@@ -122,7 +121,7 @@ export async function createNotification(
 
 			await db().insert(notifications).values({
 				id: notificationId,
-				orgId: recipientUser.activeOrganizationId,
+				orgId: undefined, // TODO: BRUHHHHH: recipientUser.activeOrganizationId,
```
oscartbeaumont, line 124: Dis needs fixing

PR #1130, opened 2025-10-04
Handles GraphicsCapture errors and initializes COM

p-delorme:
Adds more robust error handling for GraphicsCapture related operations by converting `unwrap()` calls to proper error propagation, including specific error messages to aid debugging.

Initializes COM on the thread where the capturer is created to prevent potential marshalling issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when starting screen capture on Windows, reducing crashes during initialization.
  * Prevents rare panics by gracefully handling failures when obtaining item size, creating frame pools, and starting capture sessions.
  * Provides clearer error messages when graphics capture operations fail.

* **Refactor**
  * Reworked error propagation for graphics capture operations to return user-friendly errors instead of aborting.

### DIFF
--- a/crates/scap-direct3d/src/lib.rs
+++ b/crates/scap-direct3d/src/lib.rs
@@ -38,8 +38,9 @@ use windows::{
                 IDXGIDevice,
             },
         },
-        System::WinRT::Direct3D11::{
-            CreateDirect3D11DeviceFromDXGIDevice, IDirect3DDxgiInterfaceAccess,
+        System::WinRT::{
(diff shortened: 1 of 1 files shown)


Brendonovich:
Thanks for bringing this to our attention, I made these same changes last week.
We already initialize COM outside of the capturer via `thread_init`, but reducing `unwrap` usages is pretty much always a good idea. I'd be surprised if that fixes bugs/crashes and leads to the app working properly, but if it's worked for you then that's awesome :smile:

PR #1137, opened 2025-10-06
Migrate from Remotion to Mediabunny

oscartbeaumont:
Migrate from Remotion to MediaBunny. From [here](https://www.remotion.dev/docs/media-parser/) "We are phasing out Media Parser and are moving to [Mediabunny](https://www.remotion.dev/blog/mediabunny)!"

I am testing if this will help with the supported file formats on the `UploadCapButton`.


oscartbeaumont:
This isn't going to fix the core problem and is mostly AI slop so going to close. We can piggyback on this PR if we decide this is worth it in the future.

PR #1142, opened 2025-10-06
Effect: Port `UpdateFolder`

oscartbeaumont:
Was waiting for Claude to do some stuff so thought I would start.


packages/web-backend/src/Folders/index.ts
```diff
@@ -122,6 +124,75 @@ export class Folders extends Effect.Service<Folders>()("Folders", {
...
 				yield* deleteFolder(folder);
 			}),
+
+			update: Effect.fn("Folders.update")(function* (
+				folderId: Folder.FolderId,
+				data: Folder.FolderUpdate,
+			) {
+				const user = yield* CurrentUser;
+
+				const [folder] = yield* db
```
oscartbeaumont: We lookup this `folder` soely for the 404 error. I think theorically we could use `rowsAffected` on the update later on to determine if it exists or not but that might make the logic harder to maintain.

I assume you'll wanna keep it this way though was want to check.

packages/web-backend/src/Folders/index.ts
```diff
@@ -122,6 +124,75 @@ export class Folders extends Effect.Service<Folders>()("Folders", {
...
+					.pipe(Policy.withPolicy(policy.canEdit(folderId)));
+				if (!folder) return yield* new Folder.NotFoundError();
+
+				// If parentId is provided and not null, verify it exists and belongs to the same organization
+				if (data.parentId) {
+					// Check that we're not creating a circular reference
+					if (data.parentId === folderId)
+						return yield* new Folder.RecursiveDefinitionError();
+
+					const [parentFolder] = yield* db.execute((db) =>
```
oscartbeaumont: Should this read have `policy`?

packages/web-backend/src/Folders/index.ts
```diff
@@ -122,6 +124,75 @@ export class Folders extends Effect.Service<Folders>()("Folders", {
...
+								.from(Db.folders)
+								.where(Dz.eq(Db.folders.id, currentParentId)),
+						);
+
+						if (!nextParent) break;
+						currentParentId = nextParent.parentId;
+					}
+				}
+
+				yield* db
```
oscartbeaumont: Does this need a `policy` if the folder select at the start has one which verifies it?

The policy has a DB cost to verify right so we don't wanna overuse it. We will probs add caching but I don't think that's the case at the moment?

packages/web-backend/src/Folders/index.ts
```diff
@@ -122,6 +124,75 @@ export class Folders extends Effect.Service<Folders>()("Folders", {
...
+							.from(Db.folders)
+							.where(
+								Dz.and(
+									Dz.eq(Db.folders.id, data.parentId),
+									Dz.eq(Db.folders.organizationId, user.activeOrganizationId),
+								),
+							),
+					);
+
+					if (!parentFolder) return yield* new Folder.NotFoundError();
```
oscartbeaumont: I assume this should probs be a different error right?

packages/web-domain/src/Space.ts
```diff
@@ -0,0 +1,4 @@
+import { Schema } from "effect";
+
+export const SpaceId = Schema.String; // TODO: .pipe(Schema.brand("SpaceId"));
```
oscartbeaumont, line 3: This is not branded yet for better compatibility with existing code.

I will enforce it in a follow up PR which goes through and enforces the branded types *everywhere* in the database schema.

packages/web-backend/src/Folders/FoldersPolicy.ts
```diff
@@ -1,42 +1,32 @@
...
 				Policy.policy((user) =>
 					Effect.gen(function* () {
-						const [folder] = yield* db.execute((db) =>
-							db.select().from(Db.folders).where(Dz.eq(Db.folders.id, id)),
+						const folder = yield* (yield* repo.getById(id)).pipe(
+							Effect.catchTag(
+								"NoSuchElementException",
+								() => new Policy.PolicyDeniedError(),
+							),
 						);
```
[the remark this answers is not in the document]
Brendonovich: It is correct, I don't want to increase nesting by doing a match

packages/web-backend/src/Folders/FoldersPolicy.ts
```diff
@@ -1,42 +1,32 @@
...
-									Dz.and(
-										Dz.eq(Db.spaceMembers.userId, user.id),
-										Dz.eq(Db.spaceMembers.spaceId, spaceId),
-									),
-								),
-						);
+						yield* spacesPolicy.isMember(folder.spaceId);
 
-						return spaceMember !== undefined;
+						return true;
```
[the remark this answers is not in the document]
Brendonovich: folder.spaceId` is `string | null`

PR #1269, opened 2025-10-22
Recording button race condition

oscartbeaumont:
If you spam it especially in dev it can try and start multiple recordings which absolutely wrecks the app.


oscartbeaumont:
It did look like Brendan already fixed this so gonna close!

PR #1273, opened 2025-10-22
Fix trimming behavior

### LINKED ISSUE #824: Trimming design makes little sense
### Description

I am struggling to understand how trimming works in Cap. If I create a recording then try to trim it down by dragging from the left of the timeline, the right side of the timeline shrinks down. This shouldn't be the case.

### Reproduction

Record a video, then try to trim it

### Additional Context

- Cap version: 0.3.60
- Operating system, version: macOS Sonoma

### Other Notes

Thank you for your hard work on Cap, I look forward to seeing it continue to grow!

Enejivk:
https://github.com/user-attachments/assets/d5ef9a62-e361-47b4-9762-046b6029e199

Fixed #824 

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable drag interactions for clip start/end handles with improved cleanup on blur/mouseleave and correct release behavior; prevents stale drag state.

* **Performance**
  * Smoother timeline dragging via memoized calculations that reduce recalculation and provide stable segment positioning and sizing.

* **UX**
  * Time labels on the cut-offset control now render as clean formatted text for clearer display.

### DIFF
--- a/apps/desktop/src/routes/editor/Timeline/ClipTrack.tsx
+++ b/apps/desktop/src/routes/editor/Timeline/ClipTrack.tsx
@@ -11,6 +11,7 @@ import {
 	createSignal,
 	For,
 	Match,
+	onCleanup,
 	Show,
 	Switch,
 } from "solid-js";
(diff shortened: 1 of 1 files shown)


Brendonovich:
Cleaned up some code by using 1 signal instead of one per field, which could go out of sync - better to store as little state as possible and derive other state from it.

PR #1284, opened 2025-10-23
PostHog track API backend

oscartbeaumont:
This will help us tell if all the uploader bugs are fixed because we think most of the current errors are people hosting old versions of Cap web.

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent server URL resolution at startup: debug env → stored preference → fallback env, and stored preference updated when needed.
  * Analytics now detect and label CAP Cloud vs self-hosted environments for more accurate tracking.

* **Chores**
  * Streamlined startup and settings initialization for more reliable and consistent behavior.

### DIFF
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -2238,7 +2238,7 @@ pub async fn run(recording_logging_handle: LoggingHandle, logs_dir: PathBuf) {
                     (url, false)
                 } else {
                     (
-                        optional_env!("VITE_SERVER_URL")
+                        option_env!("VITE_SERVER_URL")
                             .unwrap_or("https://cap.so")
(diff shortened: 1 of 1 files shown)


apps/desktop/src-tauri/src/lib.rs
```diff
@@ -2223,6 +2226,27 @@ pub async fn run(recording_logging_handle: LoggingHandle, logs_dir: PathBuf) {
             });
 
             {
+                let server_url = std::option_env!("VITE_SERVER_URL")
+                    .map(|s| {
+                        GeneralSettingsStore::update(&app, |state| {
```
oscartbeaumont: By updating the state the frontend will always reflect the correct URL which is useful for ensuring the setting is accurate in the UI.

PR #1286, opened 2025-10-23
Prewarm target select overlay

oscartbeaumont:
It's crazy slow on Windows so this should make it actually not suck.

## Summary by CodeRabbit

* **Improvements**
  * Target select overlays now initialize asynchronously at startup, reducing boot blocking.
  * Overlays are displayed concurrently across all connected displays for faster setup.
  * Overlays are hidden (not closed) when dismissed or during recording, improving reopen speed and preserving state.
  * Initialization errors are logged and handled gracefully to maintain stability.


oscartbeaumont:
I'm closing this in favor of merging it into #1282. These will overlap a lot and it will be easier to prevent bugs if they are done together.

PR #1338, opened 2025-11-03
web: delete org 

ameer2468:
**This PR**: adds the ability to delete an org.

## Summary by CodeRabbit

* **New Features**
  * Added ability to delete organizations with confirmation dialog.
  * System automatically switches to another organization when one is deleted.

* **Bug Fixes**
  * Deleted organizations no longer appear in dashboard queries and search results.
  * Fixed organization name text truncation on mobile layouts.

* **Style**
  * Updated destructive button styling for clarity.
  * Improved mobile organization controls layout with better spacing.
  * Adjusted settings card background colors.

### DIFF
--- a/packages/web-backend/src/Organisations/index.ts
+++ b/packages/web-backend/src/Organisations/index.ts
@@ -53,49 +53,41 @@ export class Organisations extends Effect.Service<Organisations>()(
 			) {
 				const user = yield* CurrentUser;
 
-				//this is fake deleting for now
-				yield* db
-					.use((db) =>
-						db
+				yield* Policy.withPolicy(policy.isOwner(id))(Effect.void);
+
+				// Perform tombstone, find other org, and update user in a single transaction
+				yield* db.use((db) =>
+					db.transaction(async (tx) => {
+						await tx
 							.update(Db.organizations)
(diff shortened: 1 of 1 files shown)


packages/web-domain/src/Organisation.ts
```diff
@@ -23,6 +23,11 @@ export class Organisation extends Schema.Class<Organisation>("Organisation")({
 	name: Schema.String,
 }) {}
 
+export const OrganisationDelete = Schema.Struct({
+	id: OrganisationId,
+});
+export type OrganisationDelete = Schema.Schema.Type<typeof OrganisationDelete>;
+
```
Brendonovich, line 30: This type isn't being used anywhere else, it can be inlined into the rpc payload

packages/web-backend/src/Organisations/index.ts
```diff
@@ -49,7 +48,48 @@ export class Organisations extends Effect.Service<Organisations>()(
 				}
 			});
 
-			return { update };
+			const deleteOrg = Effect.fn("Organisations.deleteOrg")(function* (
```
Brendonovich, line 51: I think this would be better named `softDelete`

PR #1627, opened 2026-02-21
feat: add disable update checks setting in the general settings

### LINKED ISSUE #1607: Setting to always ignore updates
Every single time  I open cap I am asked to upgrade, while I am 100% sure there is a bug in the new updates and have willingly downgraded to a more stable version. Can we at least stop the update checking and alerting from settings?

Starz099:
## Summary
This PR introduces a user setting that allows disabling automatic update checks.  
When enabled, the application will skip checking for updates entirely.

## Changes
- Added a new boolean field `disable_update_checks` in `GeneralSettingsStore`
- Implemented a toggle in the General Settings menu using `ToggleSettingItem`
- Updated `update.tsx` to skip update checks via an early return when the setting is enabled
- Added a dedicated signal for disabled update check state in `updates.tsx`.
- Updated `new-main/index.tsx` to skip automatic update checks on app startup when disabled

## Result
Users now have full control over whether the app performs update checks.

## Preview
image

---

image

### DIFF
--- a/apps/desktop/src-tauri/src/general_settings.rs
+++ b/apps/desktop/src-tauri/src/general_settings.rs
@@ -91,6 +91,8 @@ pub struct GeneralSettingsStore {
     #[serde(default)]
     pub hide_dock_icon: bool,
     #[serde(default)]
+    pub disable_update_checks: bool,
+    #[serde(default)]
     pub auto_create_shareable_link: bool,
     #[serde(default = "default_true")]
     pub enable_notifications: bool,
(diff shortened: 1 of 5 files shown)


richiemcilroy:
Hey! Thanks so much for the contribution. Will make a proper review soon. Just checking why you need this option? When releasing new updates our goal is to fix important bugs (sometimes critical) etc - would love to hear your thoughts here 😄

Starz099:
> Hey! Thanks so much for the contribution. Will make a proper review soon. Just checking why you need this option? When releasing new updates our goal is to fix important bugs (sometimes critical) etc - would love to hear your thoughts here 😄

This mainly comes from cases where users intentionally stay on a specific version (for stability, compatibility, or workflow reasons). Repeated update prompts can become disruptive when they’ve already decided not to upgrade.

The goal isn’t to discourage updates, but to give users control, they can still manually check whenever they choose to upgrade. Happy to adjust wording/UX if you’d prefer this framed differently.

PR #1679, opened 2026-03-25
fix: resolve camera preview and export pipeline issues

hidumou:
## Summary

This PR fixes three issues in the desktop recording/export pipeline:

1. Camera preview did not re-open when the same camera was already active.
2. Windows camera exports could show snow/noisy artifacts because camera buffers were read with the wrong stride/layout assumptions.
3. 4K 60fps export could fail with `Failed to export recording: Encode: Invalid argument` because encoded packets could reach the MP4 muxer with invalid timestamp ordering (`pts < dts`).

## Changes

- Re-open the camera preview window when the selected camera is already active but the preview window is closed.
- Respect camera buffer stride/layout on Windows during FFmpeg conversion and texture upload paths.
- Tighten the libx264 high-throughput export configuration used for 4K60 software fallback by disabling B-frames.
- Add a packet timestamp safety fix before MP4 muxing so packets are not written with `pts < dts`.

## Root Cause

For the 4K60 export failure, logs showed non-monotonic timestamp ordering during software H.264 export. The encoder/muxer path could produce packets where `pts` became smaller than `dts`, which causes FFmpeg/MP4 muxing to fail with `Invalid argument`.

For the Windows camera artifact issue, camera frames were previously treated as tightly packed buffers, but real device buffers can contain row padding/stride. That caused corrupted image interpretation in export/output paths.

## Validation

- `cargo fmt --all`
- `cargo check -p cap-enc-ffmpeg -p cap-export`
- `cargo check --manifest-path apps/desktop/src-tauri/Cargo.toml`

Additionally, the failing local 4K60 export project was re-run and progressed well past the previous early failure range without immediately reproducing `Encode: Invalid argument`.

<h3>Greptile Summary</h3>

This PR fixes three distinct bugs in the recording/export pipeline: the camera preview not re-opening when the same camera device is re-selected, Windows camera export artifacts caused by ignoring device buffer row-padding (stride), and 4K60 export failures due to `pts < dts` timestamp ordering violations.

**Key changes:**
- `lib.rs`: Adds `CapWindowId::Camera.get(&app_handle).is_none()` to the early-return branch so the preview window is re-shown only when it's actually closed; also extracts `skip_camera_window.unwrap_or(false)` to a single binding to eliminate repeated unwraps.
- `base.rs`: Introduces a secondary guard that clamps `pts` to `dts` when the encoder emits a packet with `pts < dts` independently of any DTS ordering violation — a valid scenario at stream initialisation or during B-frame drain.
- `h264.rs`: Sets `bf=0` and `b-adapt=0` in the libx264 `HighThroughput` software path, removing B-frames as the primary cause of `pts < dts` during 4K60 export.
- `win.rs`: Introduces `infer_camera_buffer_layout`, which derives `primary_stride` and `secondary_stride` from the actual buffer length, and threads both values through `copy_plane`, `flip_camera_buffer_into`, `camera_frame_to_ffmpeg`, and `upload_mf_buffer_to_texture`. All format branches (NV12, NV21, YUYV422, UYVY422, YUV420P, YV12, P010, RGB/BGR variants) now correctly skip padding bytes on both the FFmpeg conversion and D3D11 texture upload paths.

Safe to merge — all changes are targeted bug fixes with correct implementations and no regressions identified.

All three bug fixes are logically sound: the camera window guard is correct, the pts<dts clamp is a standard muxer safety net, and the stride-aware buffer handling is comprehensively applied across every format branch. The two edge-case concerns raised in prior review threads are pre-existing limitations with clear error logging, not new regressions. No P0 or P1 issues remain.

crates/recording/src/output_pipeline/win.rs — largest and most complex change; worth a final read-through of the NV21 UV swap loop and the YUV420P V-plane offset arithmetic.

| Filename | Overview |
|----------|----------|
| apps/desktop/src-tauri/src/lib.rs | Camera preview fix: adds CapWindowId::Camera.get().is_none() guard so the window is only re-opened when it's actually closed; also extracts skip_camera_window.unwrap_or(false) to a local to avoid repeated unwrap calls. |
| crates/enc-ffmpeg/src/base.rs | Adds a secondary pts<dts safety clamp after the existing DTS-monotonicity block; correctly handles the distinct case where the encoder emits pts<dts without a DTS ordering violation. |
| crates/enc-ffmpeg/src/video/h264.rs | Disables B-frames (bf=0, b-adapt=0) in the libx264 HighThroughput software fallback path, eliminating the primary source of pts<dts timestamp ordering violations during 4K60 export. |
| crates/recording/src/output_pipeline/win.rs | Large refactor introducing infer_camera_buffer_layout and compact_primary_stride to correctly handle device-supplied row padding; all format branches now use the inferred stride. Edge cases (MJPEG hard-fail in texture path, asymmetric UV stride) already noted in prior review threads. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fmt"](https://github.com/capsoftware/cap/commit/923187edfb245b4031b5e1386053191e1b967467) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26262917)</sub>

### DIFF
--- a/crates/recording/src/output_pipeline/win.rs
+++ b/crates/recording/src/output_pipeline/win.rs
@@ -1176,7 +1176,7 @@ fn infer_camera_buffer_layout(
         PixelFormat::MJPEG | PixelFormat::H264 => return None,
     };
 
-    if total_rows == 0 || data_len % total_rows != 0 {
+    if total_rows == 0 || !data_len.is_multiple_of(total_rows) {
         return None;
     }
 
@@ -1193,7 +1193,7 @@ fn infer_camera_buffer_layout(
             })
         }
         PixelFormat::YUV420P | PixelFormat::YV12 => {
-            if primary_stride < width || primary_stride % 2 != 0 {
+            if primary_stride < width || !primary_stride.is_multiple_of(2) {
                 return None;
             }
             let secondary_stride = primary_stride / 2;


richiemcilroy:
Hey @hidumou, was this an issue you have personally experienced?

richiemcilroy:
@greptileai please re-review the PR

PR #1745, opened 2026-04-21
Harden recording capture, muxing, and recovery

richiemcilroy:
This PR hardens the desktop recording pipeline so recording failures are less likely to take down the app and recovery is more reliable when a capture is interrupted.

## What changed

- adds an optional out-of-process fragmented MP4 muxer backed by a new `cap-muxer` sidecar
- bundles/builds the muxer sidecar in desktop packaging, CI, and publish workflows
- improves recovery to merge respawned fragment directories, rescue valid pending `.m4s.tmp` fragments, and quarantine corrupt fragments
- adds disk space guardrails and richer recording health signals for muxer crashes, audio degradation, device loss, encoder rebuilds, capture stalls, and target loss
- expands recording telemetry for reliability debugging and adds a desktop setting to disable anonymous telemetry
- adds new recording test coverage for recovery, muxer subprocess behavior, drift/A-V alignment, and kill-9 crash recovery scenarios

## Why

We still had cases where muxing or pipeline instability could terminate recording unexpectedly or leave behind fragmented output that was difficult to recover cleanly. This change isolates the riskiest muxing path, surfaces better health signals, and makes it much more likely that interrupted recordings can still be finalized into playable output.

## Notes

- the out-of-process muxer is exposed as an experimental reliability toggle
- if the bundled `cap-muxer` binary is unavailable, recording falls back to the in-process path
- telemetry remains anonymous and can now be disabled from Settings

<h3>Greptile Summary</h3>

This PR introduces an optional out-of-process DASH muxer sidecar (`cap-muxer`), hardens recovery to merge respawn subdirectories and rescue valid `.m4s.tmp` fragments, adds disk-space guardrails, and expands recording health telemetry. The architecture is well-thought-out — the respawning supervisor, protocol framing, and disk monitor all fit together cleanly.

- **P1 — unbounded respawn loop on disk-full**: `RespawningMuxerSubprocess::run_with_respawn` returns `DiskFull` without incrementing `consecutive_fast_failures`, so `is_exhausted()` stays `false`. The encoder loop in `oop_fragmented_m4s.rs` only sets `subprocess_fatal` when `is_exhausted()` is true, so on every subsequent frame it calls `spawn_new()` against an already-exhausted disk, looping until the 2-second `DiskSpaceMonitor` poll fires. See inline comment at `oop_muxer.rs:675`.
- **P2 — silent DASH misconfiguration**: `av_opt_set` return codes are not checked in `cap-muxer/src/main.rs`; a misspelled or unsupported option silently produces wrong segment filenames that won't match what the host expects.

Safe to merge for the in-process path; the OOP muxer experimental toggle has a P1 logic bug where disk-full causes continuous subprocess respawning for up to 2 seconds that should be fixed before shipping the feature widely.

One confirmed P1 (DiskFull doesn't set is_exhausted → respawn loop) in the experimental OOP muxer path and one P2 (silent av_opt_set failure) in the sidecar. The core recording pipeline, recovery improvements, and telemetry are solid. Score is 4 rather than 5 because the P1 is in a new code path being actively toggled on.

crates/recording/src/output_pipeline/oop_muxer.rs (DiskFull respawn loop), crates/cap-muxer/src/main.rs (av_opt_set return values, EDQUOT classification)

| Filename | Overview |
|----------|----------|
| crates/recording/src/output_pipeline/oop_muxer.rs | New out-of-process muxer substrate: spawn/respawn logic is solid, but DiskFull does not increment consecutive_fast_failures, causing an unbounded respawn loop on an exhausted disk until the 2-second disk monitor fires. |
| crates/cap-muxer/src/main.rs | New sidecar binary: async frame queue with back-pressure, correct DTS-derived duration derivation, and clean disk-full exit code; av_opt_set return values are silently ignored and EDQUOT is not classified as disk-full. |
| crates/recording/src/output_pipeline/oop_fragmented_m4s.rs | New macOS OOP pipeline: frame drop tracking, encoder rebuild on mid-stream errors, and disk monitor integration look correct; inherits the DiskFull-not-exhausted bug from oop_muxer.rs. |
| crates/recording/src/recovery.rs | Recovery now merges respawn subdirectories and rescues .m4s.tmp fragments with tail-completeness checks before renaming; single-fragment validation is now fatal for display track, which is an intentional improvement. |
| crates/utils/src/disk_space.rs | New cross-platform disk space utility using statvfs/GetDiskFreeSpaceExW with ancestor walk for non-existent paths; thresholds (50 MB stop / 200 MB warn) and tests are well-structured. |
| crates/recording/src/output_pipeline/core.rs | Adds VideoStartGate for A/V alignment, SharedHealthSender, DiskSpaceMonitor, and stall-budget send helpers; well-guarded with atomics and clear contracts. |
| apps/desktop/src-tauri/src/recording.rs | Pre-flight disk check before recording start, health event fan-out to PostHog telemetry, and RecordingHealthAccumulator integration; changes are additive and low risk. |
| apps/desktop/src-tauri/src/general_settings.rs | Adds enable_telemetry and out_of_process_muxer settings; uses OnceLock-backed set_muxer_binary_override instead of set_var to avoid the previously flagged race. |
| scripts/build-cap-muxer.sh | Clean build script that cross-compiles cap-muxer and places the target-triple-suffixed binary in the Tauri binaries directory; integrated correctly into CI and publish workflows. |
| crates/cap-muxer-protocol/src/lib.rs | New length-prefixed framing protocol with bincode serialisation for all frame types; straightforward and well-typed. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant EL as Encoder Loop
    participant RS as RespawningMuxerSubprocess
    participant MS as MuxerSubprocess
    participant CM as cap-muxer sidecar
    participant DM as DiskSpaceMonitor
    EL->>RS: write_video_packet(pkt)
    RS->>MS: write_video_packet(pkt)
    MS->>CM: stdin write
    CM-->>MS: pipe broken (disk full, exit 60)
    MS-->>RS: Err(DiskFull)
    RS->>RS: current=None, consecutive_fast_failures NOT incremented
    RS-->>EL: Err(DiskFull)
    EL->>EL: is_exhausted()? false → no subprocess_fatal
    Note over EL,DM: Next frame arrives
    EL->>RS: write_video_packet(pkt2)
    RS->>RS: current is None, not exhausted → spawn_new()
    RS->>MS: spawn [respawn-1]
    MS->>CM: launch new process
    CM-->>MS: exits 60 again
    Note over EL,DM: Loop repeats ~2s until...
    DM-->>EL: Exhausted
    EL->>EL: disk_exhausted=true
```

`````markdown
Path: crates/recording/src/output_pipeline/oop_muxer.rs
Line: 675-677

Comment:
**`DiskFull` skips fast-failure accounting, causing unbounded respawn loop on an exhausted disk**

When `run_with_respawn` receives `DiskFull` it clears `self.current` and returns immediately — without incrementing `consecutive_fast_failures`. Because `is_exhausted()` requires `consecutive_fast_failures > max_consecutive_fast_failures`, it will always return `false` after a disk-full event. The encoder thread in `oop_fragmented_m4s.rs` only sets `subprocess_fatal` when `is_exhausted()` is `true`, so on the next frame it calls `run_with_respawn` again, which sees `current == None && consecutive_fast_failures (0) <= max (3)` and calls `spawn_new()`. The freshly spawned subprocess on a full disk also dies with exit code 60, returning `DiskFull` again — repeating the cycle until the periodic disk monitor fires ~2 seconds later.

```rust
Err(MuxerSubprocessError::DiskFull(reason)) => {
    self.current = None;
+   // Mark as terminally exhausted so callers don't attempt respawn.
+   self.consecutive_fast_failures = self.max_consecutive_fast_failures + 1;
    return Err(MuxerSubprocessError::DiskFull(reason));
}
```

---

Path: crates/cap-muxer/src/main.rs
Line: 84-93

Comment:
**`classify_io_error` misses `EDQUOT` (disk quota exceeded)**

The function only tests `ENOSPC` (errno 28). On systems with per-user disk quotas, exceeding the quota produces `EDQUOT` (errno 122 on Linux, 69 on macOS). A full-quota write error won't be classified as disk-full, so the muxer exits with `EXIT_FFMPEG_ERROR` instead of `EXIT_DISK_FULL`, and the parent treats it as a crash rather than a recoverable out-of-space condition.

```rust
fn classify_io_error(err: &ffmpeg::Error) -> bool {
    let ffmpeg::Error::Other { errno } = err else { return false; };
    if *errno == libc::ENOSPC { return true; }
+   #[cfg(unix)]
+   if *errno == libc::EDQUOT { return true; }
    #[cfg(windows)]
    if *errno == 112 { return true; } // ERROR_DISK_FULL
    false
}
```

---

Path: crates/cap-muxer/src/main.rs
Line: 403-420

Comment:
**`av_opt_set` return values silently ignored for DASH configuration**

The `set_opt` closure calls `ffmpeg::ffi::av_opt_set` but discards the `c_int` return code. If any option name is unrecognised (e.g. due to a ffmpeg version change), the option is silently dropped and the muxer produces output with default segment naming — which will not match what the host process and `FragmentManifestTracker` expect. Checking the return code and propagating as an `Init` error would surface misconfiguration at startup rather than at recovery time.

---

Path: crates/recording/src/output_pipeline/oop_muxer.rs
Line: 278-286

Comment:
**Unconditional 100 ms sleep on every recording start**

`send_init_frames` sleeps for 100 ms then calls `try_wait` to detect an early subprocess exit. This hardcoded sleep fires on every OOP muxer spawn, adding fixed latency to recording start. Consider polling with a short interval (e.g. 1–5 ms) up to the 100 ms budget instead.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["clippy"](https://github.com/capsoftware/cap/commit/8d640975b1a2896e1219e761e4d64c793dedb92e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29105971)</sub>


richiemcilroy:
@greptileai please re-review the PR

richiemcilroy:
@greptileai please re-review the PR

PR #1749, opened 2026-04-22
fix(api): enforce ownership check on AI generation endpoint

MinitJain:
## Summary
- `GET /api/video/ai` authenticated the caller but never verified they own the target video
- Any authenticated user could pass `?videoId=<any_video_id>` and trigger AI generation billed to the video owner's account
- Added `eq(videos.ownerId, user.id)` to the DB query — non-owners get a 404 (same as "not found"), so the existence of other users' videos is not leaked

## Security Impact
This is an IDOR (Insecure Direct Object Reference). An attacker with a free account could exhaust another user's paid AI generation quota by repeatedly triggering generation on their videos.

## Test plan
- [ ] Owner can trigger AI generation on their own video (existing behavior unchanged)
- [ ] Authenticated non-owner gets 404 when passing another user's video ID
- [ ] Unauthenticated request still gets 401

### DIFF
--- a/apps/web/__tests__/unit/caption-tracks.test.ts
+++ b/apps/web/__tests__/unit/caption-tracks.test.ts
@@ -76,11 +76,11 @@ function createCueList(text: string): TextTrackCueList {
 function createVideo(
 	textTracks: FakeTextTrackList,
 	trackElements: FakeTrackElement[],
-): HTMLVideoElement {
+): HTMLVideoElement & FakeVideo {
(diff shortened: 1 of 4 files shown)


richiemcilroy:
Closing: already covered on main. The /api/video/ai route now gates reads through Policy.withPublicPolicy(videosPolicy.canView(videoId)) (from #1926) and scopes AI generation to the video owner. The ownership gap this PR targeted is no longer present. Thanks.

PR #1775, opened 2026-04-30
feat: add SMTP support as alternative to Resend

### LINKED ISSUE #1766: Feat: Add generic SMTP support for transactional emails
## Problem

Self-hosted Cap deployments are currently locked into Resend as the only email provider. The relevant code path is in `packages/database/emails/config.ts`, where `sendEmail()` is wired directly to the Resend SDK with no abstraction.

For self-hosters, this creates real friction:

- We're forced to add yet another third-party SaaS account just for transactional emails, even when we already operate our own SMTP infrastructure (e.g. Postal, Plunk, Listmonk, AWS SES, or a basic mail relay)
- It contradicts the spirit of self-hosting — moving data ownership away from cloud services
- Without Resend configured, magic links and team invites only appear in container logs (`sendVerificationRequest` fallback), which makes onboarding teammates impractical
- The current `accept-invite` flow effectively requires email delivery to work end-to-end

## Suggested Solution

Add a generic SMTP provider as a first-class alternative. Detection could be as simple as: if `SMTP_HOST` is set, use SMTP via `nodemailer`; otherwise fall back to Resend (or logs).

```env
SMTP_HOST=mail.example.com
SMTP_PORT=587
SMTP_USER=...
SMTP_PASSWORD=...
SMTP_FROM=auth@example.com
SMTP_SECURE=false  # or true for implicit TLS
```

The React Email rendering pipeline stays untouched — only the transport layer changes. `nodemailer` is the de-facto standard in the Node ecosystem for this and integrates cleanly with `@react-email/render`.

## Why this matters

Cap is a fantastic open-source alternative to Loom, and the self-hosting story is a big part of its appeal. SMTP support would close one of the last gaps that currently force users into specific external SaaS providers — which is exactly the lock-in self-hosters try to avoid.

julianwitzel:
## Summary

Adds SMTP transport as an alternative to Resend for sending emails (magic-link auth, notifications, etc.) on self-hosted Cap deployments. Uses `nodemailer`, which is already a transitive dependency of `@cap/database`, so no new packages are introduced.

Closes #1766.

## Why

Self-hosted Cap currently requires a Resend account for any email functionality. For users running Cap on private infrastructure, sending email through their own SMTP server (Postfix, Plunk, AWS SES via SMTP, Mailgun via SMTP, etc.) is often the natural choice — it avoids vendor lock-in, doesn't require an external account, and keeps the email path inside their own network.

This PR makes Cap's email layer pluggable: SMTP is tried first if configured, Resend is the fallback, and existing Resend deployments continue to work unchanged.

## Changes

**`packages/env/server.ts`** — Adds 6 new optional env variables:
- `SMTP_HOST`, `SMTP_PORT` (default 587), `SMTP_USER`, `SMTP_PASS`
- `SMTP_FROM` — Default From address (e.g. `Cap <cap@mail.example.com>`)
- `SMTP_SECURE` — `true` for SSL (port 465), `false` for STARTTLS (port 587)

**`packages/database/emails/config.ts`** — Extends `sendEmail()`:
- New `smtp()` helper that returns a configured `nodemailer.Transporter` when `SMTP_HOST` is set, otherwise `null`
- `sendEmail()` tries SMTP first; if no transport is configured, falls back to Resend (existing behaviour)
- React-Email templates are rendered to HTML via `@react-email/render` before being passed to nodemailer (Resend handles React directly, nodemailer needs HTML)
- From-address resolution extended: `fromOverride` → marketing → IS_CAP → `SMTP_FROM` → `RESEND_FROM_DOMAIN` → fallback

**`packages/database/auth/auth-options.ts`** — One-line fix:
```diff
-if (!serverEnv().RESEND_API_KEY) {
+if (!serverEnv().RESEND_API_KEY && !serverEnv().SMTP_HOST) {
```
Without this, the magic-link `sendVerificationRequest()` fell through to the dev-mode console-log branch even when SMTP was configured.

## Backwards compatibility

Fully backwards-compatible. When `SMTP_HOST` is not set, behaviour is identical to before — Resend handles everything as today. No migrations, no breaking changes.

## Tested

End-to-end on a self-hosted deployment:
- Magic-link auth: code email arrives in the inbox with the correct React-Email template, branding, and From address
- Verified with self-hosted Plunk SMTP (`port=587`, `secure=false`, STARTTLS)
- Confirmed Resend path still works when only `RESEND_API_KEY` is set (no SMTP env)
- Verified dev-mode console-log fallback still works when neither is configured

## Example .env

```bash
# SMTP via your own server / Plunk / SES / Mailgun
SMTP_HOST=smtp.example.com
SMTP_PORT=587
SMTP_USER=username
SMTP_PASS=secret
SMTP_FROM=Cap <cap@mail.example.com>
SMTP_SECURE=false
```

<h3>Greptile Summary</h3>

This PR adds nodemailer-based SMTP as a first-class alternative to Resend for self-hosted Cap deployments. The implementation is clean and fully backwards-compatible, but there is one present defect in `packages/database/emails/config.ts`:

- **P1 — wrong test-mode recipient over SMTP**: `to` is unconditionally set to `\"delivered@resend.dev\"` when `test: true`. That address is a Resend sandbox sink; when SMTP is the active transport it is treated as an ordinary external address and a real email is dispatched there, bypassing the intended recipient.

Safe to merge for typical self-hosted setups, but the test-mode recipient bug should be fixed before merging to avoid silently misdirected emails in any path that passes test: true with SMTP configured.

One P1 defect (Resend-specific test-sink address sent over real SMTP transport) drops the ceiling to 4/5; the issue affects a real code path and produces incorrect runtime behaviour, pulling the score to 3.

packages/database/emails/config.ts — test-mode to-address logic and from-address fallback chain.

| Filename | Overview |
|----------|----------|
| packages/database/emails/config.ts | Core email dispatch logic extended with SMTP transport; P1 bug where test-mode to-address is Resend-specific and sends real emails via SMTP, plus two P2 style concerns. |
| packages/database/auth/auth-options.ts | One-line fix correctly extends the dev-mode console-log guard to also check SMTP_HOST, preventing the fallback from triggering when only SMTP is configured. |
| packages/env/server.ts | Adds 6 well-typed optional SMTP env vars using existing Zod patterns; SMTP_PORT coercion and SMTP_SECURE boolString defaults are correct. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[sendEmail called] --> B{marketing && !IS_CAP?}
    B -- yes --> Z[return early]
    B -- no --> C[Resolve from address\nfromOverride → marketing → IS_CAP\n→ SMTP_FROM → RESEND_FROM_DOMAIN\n→ noreply@localhost]
    C --> D{test mode?}
    D -- yes --> E[to = delivered@resend.dev]
    D -- no --> F[to = email]
    E --> G{SMTP_HOST set?}
    F --> G
    G -- yes --> H[smtp transport\nnodemailer.createTransport]
    G -- no --> I{RESEND_API_KEY set?}
    H --> J[render React → HTML\ntransport.sendMail]
    I -- yes --> K[resend.emails.send\nReact passed directly]
    I -- no --> L[return Promise.resolve\nsilent no-op]
    J --> M[Email sent via SMTP]
    K --> N[Email sent via Resend]
    style E fill:#f9c,stroke:#c00
    style G fill:#ffd,stroke:#aa0
```

`````markdown

---

### Issue 1 of 3
packages/database/emails/config.ts:57
**`test` to-address is Resend-specific, sends real email via SMTP**

`"delivered@resend.dev"` is a Resend sandbox sink — it only has meaning when delivery goes through the Resend API. When SMTP is the active transport, this resolves to an ordinary external address and `sendMail()` will deliver a real email there. Any caller that passes `test: true` expecting sandboxed delivery while SMTP is configured will silently send to the wrong recipient.

```suggestion
	const to = test && !smtp() ? "delivered@resend.dev" : email;
```

### Issue 2 of 3
packages/database/emails/config.ts:52-55
**`SMTP_FROM` resolves independently of which transport is active**

The from-address chain checks `SMTP_FROM` then `RESEND_FROM_DOMAIN` before falling back to `"noreply@localhost"`. A user who configures only SMTP and never sets `SMTP_FROM` will fall through to `"noreply@localhost"`, which most SMTP servers will reject as an invalid sender, causing a silent non-delivery. Consider validating that at least one from-address env var is present when a transport is configured, or documenting this fallback clearly.

### Issue 3 of 3
packages/database/emails/config.ts:10-22
**New transporter created on every `sendEmail()` call**

`smtp()` calls `nodemailer.createTransport()` afresh for every email, discarding the internal connection pool each time. Following the memoisation pattern already used for `serverEnv()` would keep a single pooled transporter alive for the process lifetime and avoid repeated setup overhead.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat: add SMTP support as alternative to..."](https://github.com/capsoftware/cap/commit/15f27be97b9a4a9cc108f522957cbc0b77c04f95) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30357498)</sub>

> Greptile also left **3 inline comments** on this PR.

### DIFF
--- a/packages/database/auth/auth-options.ts
+++ b/packages/database/auth/auth-options.ts
@@ -73,7 +73,7 @@ export const authOptions = (): NextAuthOptions => {
 					async sendVerificationRequest({ identifier, token }) {
 						console.log("sendVerificationRequest");
 
-						if (!serverEnv().RESEND_API_KEY) {
+						if (!serverEnv().RESEND_API_KEY && !serverEnv().SMTP_HOST) {
 							console.log("\n");
 							console.log(
 								"━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━",
--- a/packages/database/emails/config.ts
+++ b/packages/database/emails/config.ts
@@ -1,9 +1,32 @@
 import { buildEnv, serverEnv } from "@cap/env";
+import { render } from "@react-email/render";
+import nodemailer from "nodemailer";
 import type { JSXElementConstructor, ReactElement } from "react";
 import { Resend } from "resend";
 
 export const resend = () =>
 	serverEnv().RESEND_API_KEY ? new Resend(serverEnv().RESEND_API_KEY) : null;
+
+let _smtpTransport: ReturnType<typeof nodemailer.createTransport> | null | undefined;
+
+export const smtp = () => {
+	if (_smtpTransport !== undefined) return _smtpTransport;
+	const env = serverEnv();
+	if (!env.SMTP_HOST) {
+		_smtpTransport = null;
+		return null;
+	}
+	_smtpTransport = nodemailer.createTransport({
+		host: env.SMTP_HOST,
+		port: env.SMTP_PORT,
+		secure: env.SMTP_SECURE,
+		auth:
+			env.SMTP_USER && env.SMTP_PASS
+				? { user: env.SMTP_USER, pass: env.SMTP_PASS }
+				: undefined,
+	});
+	return _smtpTransport;
+};
 
 export const sendEmail = async ({
 	email,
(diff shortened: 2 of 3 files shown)


julianwitzel:
Thanks for the thorough review @greptile-apps — all three points addressed in the amended commit:

- **P1** (test-mode recipient over SMTP): Now scoped to Resend only with `test && !smtp() ? "delivered@resend.dev" : email` — SMTP path uses the real recipient.
- **P2** (silent `noreply@localhost` fallback): Replaced with a descriptive `throw` so misconfigurations fail loudly at first send rather than being silently dropped by SMTP servers.
- **P2** (transporter pooling): Memoised `_smtpTransport` following the same pattern as `serverEnv()`, so the connection pool persists across calls.

KingsleyChimezie:
Thanks for the work here. The SMTP-first / Resend-fallback shape matches what we I've implemented for my private clone.
I'll get in touch with Cap team on whether #1775 or #1900 should be the landing PR for #1766. Happy to help test either path (real SMTP delivery + Resend regression) so this can move forward.

PR #1776, opened 2026-05-01
Improve desktop recording performance and reliability

richiemcilroy:
Optimizes macOS desktop recording based on MacBook Neo testing, including compatibility quality profiles, reduced camera/screen capture load, coalesced preview rendering, AVFoundation muxing fixes, decoder fallback tuning, and safer stop-recording behavior. Also refreshes the related quality/settings UI and cleans up playback, media serving, tests, and benchmark handling.

<h3>Greptile Summary</h3>

This PR optimizes macOS desktop recording performance and reliability through several interconnected changes: a new Compatibility quality tier (auto-selected for devices with < 16 GB RAM), removal of the per-frame `PixelBufferCopier` VTPixelTransferSession copy in favour of `sample_buffer.retained()`, centralized macOS QoS thread boosting via a new `cap_utils::macos_qos` module, per-request fallback distance tuning in the decoder pipeline, and `BitrateProfile`-based encoder selection. Supporting changes include AVFoundation muxing fixes (rebuild-first for timing, keyframe interval normalisation), camera-active screen-size caps, FPS limits, and TypeScript null-safety cleanups in the media server.

Safe to merge; all findings are P2 style/observability suggestions with no blocking defects.

No P0 or P1 issues found. The four P2 comments cover: FFmpeg/MF decoders silently ignoring max_fallback_distance (behavioural inconsistency but not a crash), silent FPS cap without a log, sysinfo probe on every Default call, and the dual-bool muxer config risk. Core recording logic, the new Compatibility tier, QoS boosting, and the media-server null-safety cleanup all look correct.

crates/rendering/src/decoder/ffmpeg.rs and media_foundation.rs (ignored fallback distance), crates/enc-avfoundation/src/mp4.rs (PTS cascade noted in previous review), crates/recording/src/output_pipeline/macos.rs (dual-bool config)

| Filename | Overview |
|----------|----------|
| crates/enc-avfoundation/src/mp4.rs | Adds BitrateProfile enum (Compatibility/Balanced/Ultra), PTS near-duplicate guard with minimum step, and swaps rebuild vs copy_with_new_timing order; keyframe interval now always fps (was 2×fps for normal mode) |
| crates/recording/src/output_pipeline/macos.rs | Adds compatibility_quality flag to muxer configs and centralises QoS thread boosting via cap_utils; dual-bool ultra/compatibility pattern is error-prone |
| crates/rendering/src/decoder/mod.rs | Introduces per-handle max_fallback_distance propagated through VideoDecoderMessage; default is 90 frames, camera uses 2 — but FFmpeg/MF decoders silently ignore the field |
| crates/rendering/src/decoder/avassetreader.rs | AVAssetReader now reads max_fallback_distance per request; replaces large hardcoded fallback constants with per-request values; MAX_RELAXED_FALLBACK_DISTANCE=8 used as floor for relaxed fallback scenarios |
| crates/recording/src/sources/screen_capture/macos.rs | Removes PixelBufferCopier (VTPixelTransferSession copy per frame) in favour of sample_buffer.retained(); eliminates a significant source of frame drops; uses video_info dimensions directly instead of recomputing from display scale |
| apps/desktop/src-tauri/src/general_settings.rs | Adds Compatibility quality tier; default now auto-detected from RAM (< 16 GB → Compatibility); sysinfo probe runs on each Default::default() call |
| apps/desktop/src-tauri/src/recording.rs | Caps recording FPS to 30 when camera is active (no log emitted); adds Compatibility quality branch; stop_recording now returns Ok instead of Err when no active recording |
| crates/utils/src/macos_qos.rs | New module centralising macOS QoS class helpers (UserInteractive/UserInitiated), replacing duplicate inline extern declarations |
| crates/recording/src/studio_recording.rs | Adds camera-active screen size constraints (1920×1200 normal, 1600×1000 compatibility) and disables fragmented muxing when camera is active; threads compatibility_quality to camera muxer config |
| apps/media-server/src/routes/video.ts | Replaces non-null assertions on getJob() with sendCurrentJobWebhook helper; refactors hasAudio mutable boolean to audioInput nullable object eliminating non-null assertions |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[start_recording] --> B{Camera active?}
    B -- Yes --> C[cap max_fps to 30]
    B -- No --> D[use configured max_fps]
    C --> E{Quality mode?}
    D --> E
    E -- Compatibility --> F[cap screen to 1600x1000\ncap fps to 24\nBitrateProfile::Compatibility]
    E -- Balanced --> G[cap screen to 1920x1200\nBitrateProfile::Balanced]
    E -- Ultra --> H[BitrateProfile::Ultra\nframe reordering enabled]
    F & G & H --> I[ScreenCaptureConfig::init\nwith max_capture_size]
    I --> J[AVFoundationMp4Muxer\nor AVFoundationCameraMuxer]
    J --> K[encoder threads\nboosted to UserInitiated QoS]
    K --> L[MP4Encoder::init_with_options\nBitrateProfile selects bitrate fn]
    L --> M[queue_video_frame\nPTS near-duplicate guard\nrebuild_video_sample_buf first]
```

1. `crates/enc-avfoundation/src/mp4.rs`, line 536-548 ([link](https://github.com/capsoftware/cap/blob/a9e909efec895d277fb6b74c0533f11790bbdc42/crates/enc-avfoundation/src/mp4.rs#L536-L548)) 

   <a href="#">P2</a> **PTS near-duplicate check cascades to subsequent legitimate frames**

   The new `minimum_video_pts_step()` guard (1/4 frame duration ≈ 8 ms at 30 fps) treats any incoming PTS within that window of `last_pts` as a near-duplicate and advances it by one full `frame_duration`. Once one such adjustment fires, the new `last_pts` is exactly one `frame_duration` later than expected. The very next legitimate frame — which was supposed to arrive at that same new `last_pts` — then also satisfies the condition and is advanced again, creating a permanent +1 frame cascade for the rest of the recording.

   The regression test at `regression_tiny_positive_pts_step_is_expanded` only checks that no errors occur, not that the PTS values after the second frame are correct. For a continuous recording this means all frames after the first glitch are time-shifted by ~33 ms (one frame period), which may be acceptable, but the intent should be documented. The detailed comment explaining the fallback policy was removed in the same PR, leaving the reasoning implicit.

   `````markdown
   Path: crates/enc-avfoundation/src/mp4.rs
   Line: 536-548

   Comment:
   **PTS near-duplicate check cascades to subsequent legitimate frames**

   The new `minimum_video_pts_step()` guard (1/4 frame duration ≈ 8 ms at 30 fps) treats any incoming PTS within that window of `last_pts` as a near-duplicate and advances it by one full `frame_duration`. Once one such adjustment fires, the new `last_pts` is exactly one `frame_duration` later than expected. The very next legitimate frame — which was supposed to arrive at that same new `last_pts` — then also satisfies the condition and is advanced again, creating a permanent +1 frame cascade for the rest of the recording.

   The regression test at `regression_tiny_positive_pts_step_is_expanded` only checks that no errors occur, not that the PTS values after the second frame are correct. For a continuous recording this means all frames after the first glitch are time-shifted by ~33 ms (one frame period), which may be acceptable, but the intent should be documented. The detailed comment explaining the fallback policy was removed in the same PR, leaving the reasoning implicit.

   `````
   </details>
</details>

`````markdown
Fix the following 4 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 4
crates/rendering/src/decoder/ffmpeg.rs:799
**`max_fallback_distance` silently ignored in FFmpeg and MediaFoundation decoders**

`VideoDecoderMessage::GetFrame` now carries a `max_fallback_distance` field, and `lib.rs` applies `.with_max_fallback_distance(2)` specifically to the camera decoder to keep playback from using stale frames. Both `FfmpegDecoder` and `MFDecoder` discard this field with `_`, so the tight 2-frame limit has no effect for camera segments decoded by those paths (Windows always, macOS FFmpeg fallback). The camera could serve frames up to `DEFAULT_MAX_FALLBACK_DISTANCE` (90) frames away, undermining the intent of the per-decoder setting.

### Issue 2 of 4
apps/desktop/src-tauri/src/recording.rs:134-138
**Silent FPS cap when camera is active lacks a user-visible diagnostic**

The cap to 30 fps whenever a camera feed is present is applied without any `info!` or `warn!` log. A user who configured 60 fps in settings will silently record at 30 fps with no indication in logs or UI. Adding a trace-level log (or surfacing it in the recording-started event) would make performance debugging much easier.

### Issue 3 of 4
apps/desktop/src-tauri/src/general_settings.rs:55-65
**`sysinfo` RAM probe runs on every `StudioRecordingQuality::default()` call**

`detect_total_memory_bytes()` constructs a fresh `sysinfo::System` instance (with a memory refresh) each time it is called. Because `StudioRecordingQuality`'s `Default` impl delegates here, any serde deserialization of an older settings file missing the `studio_recording_quality` key will invoke this probe. Consider caching the result with `std::sync::OnceLock` or computing it once at application start and storing it, so repeated calls don't incur the `sysinfo` overhead.

### Issue 4 of 4
crates/recording/src/output_pipeline/macos.rs:268-270
**Dual boolean flags `ultra_quality` / `compatibility_quality` can silently conflict**

`AVFoundationMp4MuxerConfig` now has two independent `bool` fields that are meant to be mutually exclusive. If both are set to `true`, the if-else chain silently picks `ultra_quality`. Replacing the two booleans with a single `BitrateProfile` (already exported from `cap_enc_avfoundation`) would make the intent explicit and eliminate the ambiguous state. `AVFoundationCameraMuxerConfig` has the same pattern.


`````

</details>

<sub>Reviews (2): Last reviewed commit: ["clippy"](https://github.com/capsoftware/cap/commit/f6a0f534d1ab6656ed637b836977b7e6c45fc264) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30419087)</sub>

### DIFF
--- a/apps/desktop/src-tauri/src/camera.rs
+++ b/apps/desktop/src-tauri/src/camera.rs
@@ -32,6 +32,14 @@ static TOOLBAR_HEIGHT: f32 = 56.0;
 
 static GPU_SURFACE_SCALE: u32 = 2;
 
+#[cfg(target_os = "macos")]
+const QOS_CLASS_USER_INTERACTIVE: u32 = 0x21;
+
+#[cfg(target_os = "macos")]
+unsafe extern "C" {
+    fn pthread_set_qos_class_self_np(qos_class: u32, relative_priority: i32) -> i32;
+}
+
 pub const MIN_CAMERA_SIZE: f32 = 150.0;
 pub const MAX_CAMERA_SIZE: f32 = 600.0;
 pub const DEFAULT_CAMERA_SIZE: f32 = 230.0;
@@ -260,6 +268,10 @@ impl CameraPreviewManager {
         });
 
         thread::spawn(move || {
+            #[cfg(target_os = "macos")]
+            unsafe {
+                pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, 0);
+            }
             LocalSet::new().block_on(
                 &rt,
                 renderer.run(window.clone(), default_state, reconfigure_rx, camera_rx),
--- a/apps/desktop/src-tauri/src/camera.rs
+++ b/apps/desktop/src-tauri/src/camera.rs
@@ -3,6 +3,8 @@ use cap_recording::{
     FFmpegVideoFrame,
     feeds::{self, camera::CameraFeed},
 };
+#[cfg(target_os = "macos")]
+use cap_utils::macos_qos::{MacOsQosClass, set_current_thread_qos};
 use ffmpeg::{
     format::{self, Pixel},
     frame,
@@ -31,14 +33,6 @@ use wgpu::{CompositeAlphaMode, SurfaceTexture};
 static TOOLBAR_HEIGHT: f32 = 56.0;
 
 static GPU_SURFACE_SCALE: u32 = 2;
-
-#[cfg(target_os = "macos")]
-const QOS_CLASS_USER_INTERACTIVE: u32 = 0x21;
-
-#[cfg(target_os = "macos")]
-unsafe extern "C" {
-    fn pthread_set_qos_class_self_np(qos_class: u32, relative_priority: i32) -> i32;
-}
 
 pub const MIN_CAMERA_SIZE: f32 = 150.0;
 pub const MAX_CAMERA_SIZE: f32 = 600.0;
@@ -269,8 +263,11 @@ impl CameraPreviewManager {
 
         thread::spawn(move || {
             #[cfg(target_os = "macos")]
-            unsafe {
-                pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, 0);
+            {
+                let result = set_current_thread_qos(MacOsQosClass::UserInteractive);
+                if result != 0 {
+                    warn!(result, "pthread_set_qos_class_self_np failed");
+                }
             }
             LocalSet::new().block_on(
                 &rt,
--- a/apps/desktop/src-tauri/src/camera_legacy.rs
+++ b/apps/desktop/src-tauri/src/camera_legacy.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU8, Ordering};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use cap_recording::FFmpegVideoFrame;
 use flume::Sender;
@@ -25,6 +25,7 @@ struct WsReadback {
 
 const WS_PREVIEW_MAX_WIDTH: u32 = 640;
 const WS_PREVIEW_MAX_HEIGHT: u32 = 360;
+const WS_PREVIEW_FRAME_INTERVAL: Duration = Duration::from_millis(33);
 
 pub async fn create_camera_preview_ws(
     blur_rx: watch::Receiver<cap_project::BackgroundBlurMode>,
@@ -40,6 +41,7 @@ pub async fn create_camera_preview_ws(
         let mut blur_rx = blur_rx;
 
         let mut blur_state = WsBlurState::new();
+        let mut next_preview_at = Instant::now();
 
         while let Ok(raw_frame) = camera_rx.recv() {
             let mut frame = raw_frame.inner;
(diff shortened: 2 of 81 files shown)


richiemcilroy:
please re-review the pr @greptileai

PR #1800, opened 2026-05-11
Monthly shareable link quota and free-plan duration limits

richiemcilroy:
Introduces a monthly cap on shareable links for free accounts and keeps the 5-minute max length enforced consistently across creation, upload, processing, and Loom import. Pro users are unchanged.

<h3>Greptile Summary</h3>

This PR enforces a 30-link monthly quota and a 5-minute maximum duration for free-plan users across every video ingestion path (desktop API, web upload, multipart, signed URL, recording-complete, media-server webhook, Loom import, and video processing workflows), while leaving Pro users unrestricted. It also ships a new `ShareableLinkUsage` dashboard widget that replaces the previous Rive animation with a live progress bar and reset date.

- **`ShareableLinkUsage.ts`** is the central enforcement module: `createVideoWithShareableLinkQuota` wraps creation in a serializable transaction with a `SELECT … FOR UPDATE` row-lock to prevent races, `assertShareableLinkDurationAllowed` handles post-creation duration checks, and `markShareableLinkUploadRejected` marks failed uploads so the quota counter excludes them.
- A composite `(ownerId, isScreenshot, createdAt)` index is added to make the monthly count query efficient, and the `UsageButton` sidebar widget is rewritten to surface real-time quota state to free users.
- Two defects need attention before merging: `VideosRepo.create` omits `isScreenshot` causing screenshots to be blocked at quota exhaustion, and the duration check in `recording-complete.ts` fails silently due to Effect error wrapping.

Not safe to merge as-is — two enforcement paths have confirmed defects that affect free-plan users in production.

The VideosRepo.create path never forwards isScreenshot, so any free user at the 30-link cap is blocked from duplicating or creating screenshots. The recording-complete endpoint wraps the duration assertion inside Effect.gen; runPromise re-throws failure as an Exit.Failure object that isShareableLinkUsageLimitError cannot detect, causing 500 responses instead of 403 upgrade prompts for over-length segment recordings.

packages/web-backend/src/Videos/VideosRepo.ts (missing isScreenshot forwarding) and apps/web/app/api/upload/[...route]/recording-complete.ts (Effect error wrapping mismatch).

| Filename | Overview |
|----------|----------|
| packages/web-backend/src/ShareableLinkUsage.ts | New module implementing the monthly quota and duration limit logic; core functions are well-structured with row-locking in assertCanCreateShareableLink. |
| packages/web-backend/src/Videos/VideosRepo.ts | createVideoWithShareableLinkQuota is integrated but isScreenshot is never forwarded from data, causing screenshot creation to be incorrectly blocked once a free user's quota is exhausted. |
| apps/web/app/api/upload/[...route]/recording-complete.ts | Duration check is placed inside Effect.gen; when it fails the outer catch receives an Exit.Failure object that isShareableLinkUsageLimitError cannot recognize, resulting in 500 instead of 403. |
| apps/web/app/api/upload/[...route]/multipart.ts | Duration check correctly handled inside the Effect pipeline via catchAll; isScreenshot forwarded from the fetched video record. |
| apps/web/app/api/upload/[...route]/signed.ts | Duration assertion uses plain async/await before any Effect code; markShareableLinkUploadRejected correctly called on limit error. |
| apps/web/app/api/desktop/[...route]/video.ts | Replaces the old inline userIsPro + duration check with createVideoWithShareableLinkQuota; also adds a missing ownership guard on the existing-video path. |
| apps/web/workflows/process-video.ts | Duration check added with correct async/await pattern; markShareableLinkUploadRejected called before re-throwing. |
| apps/web/workflows/import-loom-video.ts | Duration check mirrors process-video.ts; createVideoWithShareableLinkQuota correctly wraps the three-table insert in a single transaction. |
| packages/database/schema.ts | Adds composite index on (ownerId, isScreenshot, createdAt) to support the monthly count query; migration file matches. |
| apps/web/components/UsageButton.tsx | Replaces the Rive animation button with a clean usage widget showing links used, progress bar, and reset date. |
| apps/web/app/api/webhooks/media-server/progress/route.ts | Duration check inserted before the metadata update; returns HTTP 200 to the media server on limit error while still marking the upload as rejected. |
| apps/web/__tests__/unit/shareable-link-usage.test.ts | New unit tests covering period boundaries, usage snapshot formatting, allow/block thresholds, and wrapped-error detection. |

</details>

1. `packages/web-backend/src/Videos/VideosRepo.ts`, line 65-120 ([link](https://github.com/capsoftware/cap/blob/ccf2838566c08da71ea99148831fd08775273422/packages/web-backend/src/Videos/VideosRepo.ts#L65-L120)) 

   <a href="#">P1</a> **`isScreenshot` not forwarded — screenshot creation blocked when quota is exhausted**

   `createVideoWithShareableLinkQuota` is called without `isScreenshot`, so it defaults to `false`. When a free user has reached the 30-link cap, `assertCanCreateShareableLink` throws `ShareableLinkUsageLimitError` for every call through this path — including for screenshot videos — because `isScreenshot` is always treated as `false`. Screenshots should be unconditionally allowed (they are excluded from `countCurrentPeriodShareableLinks`), but any `VideosRepo.create` invocation (e.g. `VideoDuplicate`, `VideoInstantCreate`) that carries `data.isScreenshot === true` will be incorrectly rejected once the user's quota is full.

   The fix is to read `isScreenshot` from `data` and forward it: `isScreenshot: data.isScreenshot ?? false`.

   `````markdown
   Path: packages/web-backend/src/Videos/VideosRepo.ts
   Line: 65-120

   Comment:
   **`isScreenshot` not forwarded — screenshot creation blocked when quota is exhausted**

   `createVideoWithShareableLinkQuota` is called without `isScreenshot`, so it defaults to `false`. When a free user has reached the 30-link cap, `assertCanCreateShareableLink` throws `ShareableLinkUsageLimitError` for every call through this path — including for screenshot videos — because `isScreenshot` is always treated as `false`. Screenshots should be unconditionally allowed (they are excluded from `countCurrentPeriodShareableLinks`), but any `VideosRepo.create` invocation (e.g. `VideoDuplicate`, `VideoInstantCreate`) that carries `data.isScreenshot === true` will be incorrectly rejected once the user's quota is full.

   The fix is to read `isScreenshot` from `data` and forward it: `isScreenshot: data.isScreenshot ?? false`.

   `````
   </details>

</details>

`````markdown

---

### Issue 1 of 2
apps/web/app/api/upload/[...route]/recording-complete.ts:124-133
**Duration limit error not detected in outer catch block**

`assertShareableLinkDurationAllowed` is called inside an `Effect.gen` block and its failure propagates as an `Exit.Failure` object (thrown by the custom `runPromise` via `throw res`). The outer `catch (error)` block then calls `isShareableLinkUsageLimitError(error)`, but that function only checks `error._tag` and recursively checks `error.cause` on `Error` instances — `Exit.Failure` is a plain Effect data object (not an `Error` instance) with `_tag: "Failure"`, so the check returns `false`. The result is a 500 response instead of a 403 with an upgrade prompt, and `markShareableLinkUploadRejected` is never called via the intended path.

Compare with `multipart.ts`, where the error is handled *inside* the Effect chain using `Effect.catchAll`, so `isShareableLinkUsageLimitError` receives the raw error rather than a wrapped `Exit.Failure`. The simplest fix here is to move the duration assertion before the `Effect.gen` block (using plain `async/await` like `signed.ts` does), or to catch it inside the Effect pipeline the same way `multipart.ts` does.

### Issue 2 of 2
packages/web-backend/src/Videos/VideosRepo.ts:65-120
**`isScreenshot` not forwarded — screenshot creation blocked when quota is exhausted**

`createVideoWithShareableLinkQuota` is called without `isScreenshot`, so it defaults to `false`. When a free user has reached the 30-link cap, `assertCanCreateShareableLink` throws `ShareableLinkUsageLimitError` for every call through this path — including for screenshot videos — because `isScreenshot` is always treated as `false`. Screenshots should be unconditionally allowed (they are excluded from `countCurrentPeriodShareableLinks`), but any `VideosRepo.create` invocation (e.g. `VideoDuplicate`, `VideoInstantCreate`) that carries `data.isScreenshot === true` will be incorrectly rejected once the user's quota is full.

The fix is to read `isScreenshot` from `data` and forward it: `isScreenshot: data.isScreenshot ?? false`.


`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(upload): mark quota rejects on sign..."](https://github.com/capsoftware/cap/commit/ccf2838566c08da71ea99148831fd08775273422) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31659259)</sub>


richiemcilroy:
please re-review the pr @greptileai - added support for older desktop versions.

richiemcilroy:
please re-review the pr @greptileai- added support for older desktop versions.

PR #1831, opened 2026-05-17
perf(recording): reduce recording memory pressure

ibbxx:
/claim #109

## Summary
- Stream cursor flushes directly to disk without cloning and pretty-printing the full cursor event. history every interval.
- - Skip periodic cursor flushes when no cursor move/click events were added since the last write.
- - Lower instant-mode MP4 muxer queue default from 240 to 96 frames and add CAP_MP4_MUXER_BUFFER_SIZE_INSTANT for instant-only tuning.
- - Serialize muxer env-var tests behind a shared lock so process-wide env mutation does not race under parallel cargo test.
## Why this helps
Long recordings currently reallocateand serialize the entire cursor history on every flush tick, even when no cursor event count changes. Streaming writes and no-op flush skipping reduce duration-scaled allocations. Lowering the instant muxer queue reduces retained frame capacity under encode/upload backpressure while keeping the global override available and adding an instant-specific override.

<h3>Greptile Summary</h3>

This PR reduces recording memory pressure through two independent changes: streaming cursor data directly to disk with borrowed slices instead of cloning and pretty-printing the full history, and lowering the instant-mode MP4 muxer frame queue from 240 to 96 with a new `CAP_MP4_MUXER_BUFFER_SIZE_INSTANT` env override.

- **cursor.rs**: `flush_cursor_data` now uses `File::create` + `BufWriter` + `serde_json::to_writer` with a `BorrowedCursorEvents<'a>` wrapper to avoid allocations; a length-comparison guard skips the flush tick when no new cursor events have been recorded since the last write.
- **macos.rs**: `get_mp4_muxer_buffer_size` gains a two-level env-var lookup (instant-specific override → global override → compile-time default); existing env-mutation tests are refactored behind a shared `ENV_LOCK` mutex but the helper `with_muxer_env` is not panic-safe, which can cause mutex poisoning and cascade-fail the test module.

The production recording changes are straightforward and low-risk, but the test refactor in macos.rs introduces a fragility that can make the entire test module fail as a cascade from a single assertion error.

The cursor streaming and muxer queue changes are correct. The concern is the with_muxer_env helper: when any test closure panics, Rust poisons the ENV_LOCK mutex, making every subsequent lock().unwrap() call panic — turning one test failure into a whole-module failure. A secondary issue is that File::create truncates the cursor output file before serialization completes, leaving an empty file on I/O error rather than preserving the last valid snapshot.

crates/recording/src/output_pipeline/macos.rs — specifically the with_muxer_env test helper and its mutex-poisoning exposure on panic.

| Filename | Overview |
|----------|----------|
| crates/recording/src/cursor.rs | Streaming write optimization with BorrowedCursorEvents avoids cloning; skip-flush guard uses length-based comparison; File::create now truncates before serialization, risking data loss on write error. |
| crates/recording/src/output_pipeline/macos.rs | Instant-mode MP4 muxer queue lowered from 240→96; new CAP_MP4_MUXER_BUFFER_SIZE_INSTANT env var added with correct fallback logic; test helper with_muxer_env introduced but not panic-safe — mutex poisoning on test failure will cascade to all sibling tests. |

</details>

1. `crates/recording/src/cursor.rs`, line 62-88 ([link](https://github.com/capsoftware/cap/blob/63eea3472093d499376a6d0dbd11aa55fb7305bb/crates/recording/src/cursor.rs#L62-L88)) 

   <a href="#">P2</a> **File truncated before serialization succeeds**

   `File::create` opens and truncates the output file before `serde_json::to_writer` has written anything. If serialization fails (e.g. mid-stream I/O error on the writer), the file is left empty or with partial bytes flushed on `BufWriter` drop — permanently losing the previously valid cursor data. The old code built the full JSON string in memory first, so the file was not touched on serialization failure. A safer pattern is to write to a temporary file and rename it over the destination only after a successful flush.

   `````markdown
   Path: crates/recording/src/cursor.rs
   Line: 62-88

   Comment:
   **File truncated before serialization succeeds**

   `File::create` opens and truncates the output file before `serde_json::to_writer` has written anything. If serialization fails (e.g. mid-stream I/O error on the writer), the file is left empty or with partial bytes flushed on `BufWriter` drop — permanently losing the previously valid cursor data. The old code built the full JSON string in memory first, so the file was not touched on serialization failure. A safer pattern is to write to a temporary file and rename it over the destination only after a successful flush.

   `````
   </details>

</details>

`````markdown

---

### Issue 1 of 3
crates/recording/src/output_pipeline/macos.rs:1207-1215
**Mutex poisoning cascades on test failure**

If any test panics inside `with_muxer_env` (e.g. an `assert_eq!` fails), the thread unwinds while holding `_guard`. Rust marks the `Mutex` as poisoned when a thread panics while holding a lock. Every subsequent call to `ENV_LOCK.lock().unwrap()` will then panic with a `PoisonError`, causing all remaining tests in the module to fail regardless of their own correctness. Use `.unwrap_or_else(|e| e.into_inner())` to tolerate a poisoned mutex.

### Issue 2 of 3
crates/recording/src/cursor.rs:62-88
**File truncated before serialization succeeds**

`File::create` opens and truncates the output file before `serde_json::to_writer` has written anything. If serialization fails (e.g. mid-stream I/O error on the writer), the file is left empty or with partial bytes flushed on `BufWriter` drop — permanently losing the previously valid cursor data. The old code built the full JSON string in memory first, so the file was not touched on serialization failure. A safer pattern is to write to a temporary file and rename it over the destination only after a successful flush.

### Issue 3 of 3
crates/recording/src/output_pipeline/macos.rs:1229-1237
Env vars set inside `with_muxer_env` are not cleaned up if `test()` panics. After the panic is caught by the test harness, `CAP_MP4_MUXER_BUFFER_SIZE` / `CAP_MP4_MUXER_BUFFER_SIZE_INSTANT` remain set, which can pollute later tests. Wrapping `test()` in `catch_unwind` and resuming after cleanup ensures the vars are always removed.

```suggestion
            let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(test));

            unsafe {
                std::env::remove_var("CAP_MP4_MUXER_BUFFER_SIZE");
                std::env::remove_var("CAP_MP4_MUXER_BUFFER_SIZE_INSTANT");
            }

            match result {
                Ok(v) => v,
                Err(e) => std::panic::resume_unwind(e),
            }
        }
```


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["perf(recording): reduce recording memory..."](https://github.com/capsoftware/cap/commit/63eea3472093d499376a6d0dbd11aa55fb7305bb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32475857)</sub>

### DIFF
--- a/crates/recording/src/cursor.rs
+++ b/crates/recording/src/cursor.rs
@@ -7,7 +7,7 @@ use cap_timestamp::Timestamps;
 use futures::{FutureExt, future::Shared};
 use std::{
     collections::HashMap,
-    fs::File,
+    fs::{self, File},
     io::{BufWriter, Write},
     path::{Path, PathBuf},
     time::Instant,
@@ -59,11 +59,20 @@ fn flush_cursor_data(output_path: &Path, moves: &[CursorMoveEvent], clicks: &[Cu
         moves: &'a [CursorMoveEvent],
     }
 
-    let file = match File::create(output_path) {
+    let temp_path = output_path.with_extension(
+        output_path
+            .extension()
+            .and_then(|ext| ext.to_str())
+            .map(|ext| format!("{ext}.tmp"))
+            .unwrap_or_else(|| "tmp".to_string()),
+    );
+
+    let file = match File::create(&temp_path) {
         Ok(file) => file,
         Err(e) => {
             tracing::error!(
-                "Failed to create cursor data file {}: {}",
+                "Failed to create temporary cursor data file {} for {}: {}",
+                temp_path.display(),
                 output_path.display(),
                 e
             );
@@ -75,16 +84,36 @@ fn flush_cursor_data(output_path: &Path, moves: &[CursorMoveEvent], clicks: &[Cu
     let mut writer = BufWriter::new(file);
     if let Err(e) = serde_json::to_writer(&mut writer, &events) {
         tracing::error!(
-            "Failed to serialize cursor data to {}: {}",
+            "Failed to serialize cursor data to temporary file {} for {}: {}",
+            temp_path.display(),
             output_path.display(),
             e
         );
-    } else if let Err(e) = writer.write_all(b"\n").and_then(|_| writer.flush()) {
+        let _ = fs::remove_file(&temp_path);
+        return;
+    }
+
+    if let Err(e) = writer.write_all(b"\n").and_then(|_| writer.flush()) {
         tracing::error!(
-            "Failed to flush cursor data to {}: {}",
+            "Failed to flush cursor data to temporary file {} for {}: {}",
+            temp_path.display(),
             output_path.display(),
             e
         );
+        let _ = fs::remove_file(&temp_path);
+        return;
+    }
+
+    drop(writer);
+
+    if let Err(e) = fs::rename(&temp_path, output_path) {
+        tracing::error!(
+            "Failed to replace cursor data file {} with temporary file {}: {}",
+            output_path.display(),
+            temp_path.display(),
+            e
+        );
+        let _ = fs::remove_file(&temp_path);
     }
 }
 
--- a/crates/recording/src/cursor.rs
+++ b/crates/recording/src/cursor.rs
@@ -1,12 +1,14 @@
 use cap_cursor_capture::CursorCropBounds;
 use cap_cursor_info::CursorShape;
 use cap_project::{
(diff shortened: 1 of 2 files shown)


ibbxx:
Fixed in the latest commit. The test helper now tolerates poisoned env mutexes and always cleans up the muxer env vars before resuming any panic.

PR #1630, opened 2026-02-24
fix: bypass broken workflow package for self-hosted transcription and AI generation

oaris-dev:
## Summary

Fixes self-hosted transcription and AI generation by bypassing the broken `workflow` package `[local world]` mode. Resolves #1550.

- **`lib/transcribe.ts`**: Replace `start(transcribeVideoWorkflow, ...)` with direct `transcribeVideoDirect()` that validates, extracts audio, calls Deepgram, saves VTT, and cleans up
- **`lib/generate-ai.ts`**: Replace `start(generateAiWorkflow, ...)` with direct `generateAiDirect()` that fetches transcript, calls AI APIs (Groq → OpenAI fallback), and saves metadata
- **`actions/videos/get-status.ts`**: Set PROCESSING in DB before firing transcription (prevents re-trigger loops on every 2s poll), add 3-minute stale PROCESSING → ERROR timeout
- **HomePage components**: Fix TypeScript strict mode errors in 4 files (unrelated, split into separate commit)

## Context

The `workflow` package (`4.0.1-beta.42`) `[local world]` mode crashes with `TypeError: Cannot perform ArrayBuffer.prototype.slice on a detached ArrayBuffer` on all Node versions (20, 22, 24). This breaks transcription and AI generation for **every** self-hosted Docker deployment (7+ users confirmed in #1550).

The root cause: `start()` resolves successfully, but the workflow crashes asynchronously in the background queue. The catch block in the caller never fires, so `transcriptionStatus` stays null → the 2-second polling loop re-triggers transcription indefinitely.

Cap Cloud uses a distributed workflow runner (`web-cluster` via `WORKFLOWS_RPC_URL`) which doesn't have this issue. The workflow files are preserved so Cap Cloud continues to work unchanged.

We've tested this on a self-hosted staging instance (`ghcr.io/oaris-dev/cap-web:staging`) with Deepgram + OpenAI configured — transcription, AI generation, NO_AUDIO detection, and stale timeout all work correctly.

## Test plan

- [ ] Deploy self-hosted Docker image with `DEEPGRAM_API_KEY` configured
- [ ] Upload a video with audio → verify transcription completes (VTT generated, status COMPLETE)
- [ ] Verify AI generation completes if `GROQ_API_KEY` or `OPENAI_API_KEY` is set
- [ ] Upload a video without audio → verify NO_AUDIO status (not stuck at PROCESSING)
- [ ] Verify stale PROCESSING timeout triggers after 3 minutes

[...]

PR #1832, opened 2026-05-17
fix(web): write PROCESSING status to DB before triggering transcription workflow

### LINKED ISSUE #1550: Transcription workflow fails on Docker image with Node 24
### Description

  

  Self-hosted Docker deployment fails to run transcription workflows. The workflow queue crashes immediately with an ArrayBuffer error before reaching Deepgram or any actual transcription logic.

  **Error from logs:**
  [local world] Queue operation failed: TypeError: fetch failed
      at async (.next/server/chunks/2242.js:30:10424) {
    [cause]: TypeError: Cannot perform ArrayBuffer.prototype.slice on a detached ArrayBuffer
        at ArrayBuffer.slice ()
  }

  **Expected behavior:** Transcription should process and generate captions for videos.

  **Root cause analysis:** The `workflow` package (`4.0.1-beta.42`) appears incompatible with Node 24. The current Dockerfile uses `node:24-alpine`, and the ArrayBuffer detachment error is consistent with V8
  changes in newer Node versions affecting buffer handling in the workflow package's local execution mode.

  ### Reproduction

  

  1. Deploy `ghcr.io/capsoftware/cap-web:latest` via Docker (using Coolify)
  2. Set `DEEPGRAM_API_KEY` environment variable
  3. Upload a video with audio
  4. View the video share page
  5. Observe repeated `[local world] Queue operation failed` errors in logs

  ### Additional Context

  - Cap version: `ghcr.io/capsoftware/cap-web:latest`
  - Operating system, version: Docker on Coolify (Linux host)
  - Device (optional): Self-hosted server

  **Container Node version:** `v24.12.0`

  **Potential fix???** Downgrade to Node 20 or 22 LTS in the Dockerfile, or update the `workflow` package to a version compatible with Node 24.

oaris-dev:
## Summary

Fixes #1550. Writes `transcriptionStatus = 'PROCESSING'` to the DB synchronously **after** `transcribeVideo()`'s existing early-return check but **before** the `start(transcribeVideoWorkflow, ...)` call. Prevents the share page's 2-second polling loop from re-firing `start()` concurrently, which races `@workflow/world-local`'s queue → ArrayBuffer detachment → queue crash before any step executes.

## Root cause

The `/s/<videoId>` page polls `getVideoStatus()` every 2 seconds. Upstream's flow:

1. Poll 1: `video.transcriptionStatus === null` → calls `transcribeVideo()` → calls `start(transcribeVideoWorkflow, ...)` → workflow dispatched asynchronously
2. Poll 2 (2s later): `video.transcriptionStatus === null` *still* (the workflow's `validateVideo` step hasn't written PROCESSING yet) → calls `transcribeVideo()` again → calls `start()` *again*
3. Repeat...

After a few concurrent `start()` calls, `@workflow/world-local`'s queue races on its internal buffer handling, the underlying `ArrayBuffer` gets detached, and dispatch crashes with `TypeError: Cannot perform ArrayBuffer.prototype.slice on a detached ArrayBuffer` *before* any workflow step runs. The crash happens upstream of the step boundary, which is why affected users see no `/audio/extract` calls in their media-server logs (per @julianwitzel's investigation in #1550).

## The fix

`transcribeVideo()` already has an early-return guard for `transcriptionStatus === 'PROCESSING'` (lines 89–99). The problem is that nothing writes PROCESSING synchronously — the workflow's own step writes it, but only *after* the queue dispatches. If the queue race fires before the step runs, PROCESSING never gets written and the next poll triggers another `start()`.

By writing PROCESSING to the DB inside `transcribeVideo()` right before the `start()` call, the next poll's call hits the early-return at line 89 and never re-fires `start()`. The first dispatch completes normally.

5-line change in `apps/web/lib/transcribe.ts`.

## Verification

- @julianwitzel verified end-to-end on a self-hosted Docker deployment that was previously stuck in the ArrayBuffer crash loop. Logs after the patch:
  - `[transcribe] Probe result: audioCodec=aac, audioChannels=1, sampleRate=48000` ✓
  - `[transcribe] Extracted audio: 179860 bytes` ✓ (first time `/audio/extract` ever succeeded in their setup)
  - `transcriptionStatus → COMPLETE` ✓
  - AI summary + chapters generated correctly via Groq ✓
  - No more `[local world] Queue operation failed: ArrayBuffer detached` loop

## Test plan

- [x] Self-hosted Docker deployment with `DEEPGRAM_API_KEY` configured — verified by @julianwitzel
- [ ] Upload a fresh video, watch `/s/<videoId>` page, confirm `transcriptionStatus` transitions `null` → `PROCESSING` → `COMPLETE` exactly once (no re-trigger loop)
- [ ] Confirm media-server logs show `/video/probe` and `/audio/extract` calls
- [ ] Confirm no ArrayBuffer errors in `cap-web` logs

## Notes

This supersedes #1630 (which proposed bypassing the workflow entirely for transcription/AI generation). The bypass approach was overkill — this 5-line guard is the actual minimal fix.

<h3>Greptile Summary</h3>

This PR fixes a concurrency bug where the share page's 2-second polling loop could fire `transcribeVideo()` multiple times before the workflow's own first step wrote `PROCESSING`, causing concurrent `start()` calls that raced `@workflow/world-local`'s internal buffer and crashed with an `ArrayBuffer` detachment error. The fix writes `transcriptionStatus = 'PROCESSING'` to the DB synchronously inside `transcribeVideo()`, just before the `start()` call, so subsequent polls hit the existing early-return guard and never re-dispatch.

- **5-line addition** in `apps/web/lib/transcribe.ts`: persists `PROCESSING` to the DB before `start(transcribeVideoWorkflow, …)`, and the existing `catch` block already resets the status back to `null` if `start()` throws, preserving error-recovery behavior.
- A very small TOCTOU window remains (two requests could both pass the guard before either writes PROCESSING), but in practice it is negligible at a 2-second poll interval and far narrower than the pre-fix window that spanned the entire workflow execution.

The change is minimal and well-reasoned; the existing catch block correctly resets status on failure, so the main risk is the narrow TOCTOU gap that is practically negligible at normal polling intervals.

The fix correctly targets the root cause and error recovery is preserved. The only remaining concern is a theoretical race that requires two requests to overlap within a single DB round-trip, which is unlikely at a 2-second polling interval but could surface if the interval is reduced or DB latency spikes.

apps/web/lib/transcribe.ts — the only changed file; review the interaction between the new PROCESSING write and the catch block's null reset if `start()` fails.

| Filename | Overview |
|----------|----------|
| apps/web/lib/transcribe.ts | Adds a synchronous DB write of `transcriptionStatus = 'PROCESSING'` before calling `start(transcribeVideoWorkflow, ...)`, preventing re-entrant workflow dispatches from the share page's 2-second poll. The existing catch block already resets the status to `null` on failure, so error recovery is handled. A very narrow TOCTOU window remains (two requests reading the video before either writes PROCESSING), but it is practically negligible at a 2-second polling interval. |

</details>

1. `apps/web/lib/transcribe.ts`, line 118-129 ([link](https://github.com/capsoftware/cap/blob/74627a69e6e2c8f7ddf047696e562c642134789a/apps/web/lib/transcribe.ts#L118-L129)) 

   <a href="#">P2</a> **Narrow TOCTOU window still exists between guard read and PROCESSING write**

   Two concurrent poll requests can both pass the `transcriptionStatus === 'PROCESSING'` early-return guard (lines 84–93) before either one writes PROCESSING to the DB — the guard uses the value fetched at the top of the function, not a fresh read. The race window is now just the upload-phase query (~one round-trip), versus the entire workflow execution time before this fix, so it's highly unlikely to fire in practice at a 2-second polling interval. Worth noting in case the polling interval is ever reduced or the DB write is slow under load.

   `````markdown
   Path: apps/web/lib/transcribe.ts
   Line: 118-129

   Comment:
   **Narrow TOCTOU window still exists between guard read and PROCESSING write**

   Two concurrent poll requests can both pass the `transcriptionStatus === 'PROCESSING'` early-return guard (lines 84–93) before either one writes PROCESSING to the DB — the guard uses the value fetched at the top of the function, not a fresh read. The race window is now just the upload-phase query (~one round-trip), versus the entire workflow execution time before this fix, so it's highly unlikely to fire in practice at a 2-second polling interval. Worth noting in case the polling interval is ever reduced or the DB write is slow under load.

   `````
   </details>

</details>

`````markdown

---

### Issue 1 of 1
apps/web/lib/transcribe.ts:118-129
**Narrow TOCTOU window still exists between guard read and PROCESSING write**

Two concurrent poll requests can both pass the `transcriptionStatus === 'PROCESSING'` early-return guard (lines 84–93) before either one writes PROCESSING to the DB — the guard uses the value fetched at the top of the function, not a fresh read. The race window is now just the upload-phase query (~one round-trip), versus the entire workflow execution time before this fix, so it's highly unlikely to fire in practice at a 2-second polling interval. Worth noting in case the polling interval is ever reduced or the DB write is slow under load.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): write PROCESSING status to DB ..."](https://github.com/capsoftware/cap/commit/74627a69e6e2c8f7ddf047696e562c642134789a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32501118)</sub>

### DIFF
--- a/apps/web/lib/transcribe.ts
+++ b/apps/web/lib/transcribe.ts
@@ -115,6 +115,11 @@ export async function transcribeVideo(
 			`[transcribeVideo] Triggering transcription workflow for video ${videoId}`,
 		);
 
+		await db()
+			.update(videos)
+			.set({ transcriptionStatus: "PROCESSING" })
+			.where(eq(videos.id, videoId));
+
 		await start(transcribeVideoWorkflow, [
 			{
 				videoId,


oaris-dev:
FYI — opened #1833 alongside this PR. When end-to-end validating this fix against a clean upstream build, I found a second blocker on self-hosted Docker: `proxy.ts` redirects `/.well-known/workflow/v1/*` (used by the workflow queue's self-dispatch) to `/login`, so the workflow engine can't run at all.

Both fixes are needed for self-hosted transcription to actually work end-to-end. They're independent code paths in different files, so I split them into two PRs for clarity, but happy to fold into one if you prefer.

oaris-dev:
Closing this — it has been fixed upstream independently in 08990bb90 ("fix(web): prevent duplicate transcription triggers with atomic claim").

For anyone finding this later: that commit is the better fix. It performs an atomic claim — a conditional `UPDATE ... WHERE transcriptionStatus IS NULL` plus an affected-rows check — where this PR only wrote the status unconditionally before dispatch. The conditional form closes the narrow within-poll race that this PR left open.

We have verified against current `main` that self-hosted transcription completes end to end with 08990bb90 in place, and have dropped our local patch during our sync to v0.4.87. Thanks to whoever picked it up.

PR #1849, opened 2026-05-20
fix(macos): remove redundant activation policy toggle causing duplicate dock icons

ManthanNimodiya:
## What does this PR fix?
Fixes the bug where multiple Cap app icons appear in the macOS Dock when starting a recording.

## Root cause
`prepare_macos_panel_window` eagerly called `set_activation_policy(Accessory)` on every panel window creation. During recording start, multiple panel windows are created in sequence, causing rapid `Accessory → Regular` policy toggling that confuses macOS Dock into registering duplicate app tiles.

## Fix
Removed the eager `set_activation_policy(Accessory)` call — the existing `MACOS_PENDING_PANEL_WINDOWS` counter already blocks the dock sync from firing during panel creation, making it redundant.

## Changes
- `apps/desktop/src-tauri/src/permissions.rs`: removed 4 lines from `prepare_macos_panel_window`

<h3>Greptile Summary</h3>

This PR adds two cooperative guards in `permissions.rs` to prevent duplicate Dock icons that appeared when multiple panel windows were opened during recording start. The root cause was a race window where `MACOS_PENDING_PANEL_WINDOWS` could briefly reach zero between sequential window creations, letting a deferred `sync_macos_dock_visibility` run and flip the activation policy back to `Regular` before the next `prepare_macos_panel_window` call set it to `Accessory` again.

- `prepare_macos_panel_window` now calls `set_activation_policy(Accessory)` only when transitioning from 0→1 pending windows (`prev == 0`), eliminating redundant policy writes for every subsequent panel window in a batch.
- `sync_macos_dock_visibility` gains a new early-return that skips dock synchronization whenever any panel window is currently visible — a second line of defense that catches the brief-counter-zero race the pending counter alone couldn't block.

Safe to merge — both changes are narrowly scoped to macOS dock-state management and cannot regress on non-macOS targets.

The two changes are tightly complementary and each independently correct: the 0→1 guard prevents redundant Accessory policy writes, and the visible-panel-window early-return closes the brief-zero-counter race that was the actual source of duplicate Dock tiles. The existing atomic counter and 100ms debounced sync remain intact and interact correctly with both new guards. No data races, no missing error paths, and the fallback in the visibility check errs conservatively.

| Filename | Overview |
|----------|----------|
| apps/desktop/src-tauri/src/permissions.rs | Two complementary guards added: `set_activation_policy(Accessory)` now only fires on the 0→1 transition, and `sync_macos_dock_visibility` bails early when any panel window is still visible — both correct and coherent with the existing atomic counter logic. |

</details>

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;fix/macos-duplicate-dock-i..."](https://github.com/capsoftware/cap/commit/7b0615d48f9ec574baac8dede08715895bca7160) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33137642)</sub>

### DIFF
--- a/apps/desktop/src-tauri/src/permissions.rs
+++ b/apps/desktop/src-tauri/src/permissions.rs
@@ -162,7 +162,13 @@ fn macos_sync_activation_policy(app: &tauri::AppHandle, should_show_dock: bool) 
 pub(crate) fn prepare_macos_panel_window(
     app: &tauri::AppHandle,
 ) -> MacosPanelWindowActivationGuard {
-    MACOS_PENDING_PANEL_WINDOWS.fetch_add(1, Ordering::AcqRel);
+    let prev = MACOS_PENDING_PANEL_WINDOWS.fetch_add(1, Ordering::AcqRel);
+
+    if prev == 0 {
+        if let Err(err) = app.set_activation_policy(tauri::ActivationPolicy::Accessory) {
+            tracing::warn!("Failed to prepare macOS panel activation policy: {err}");
+        }
+    }
 
     MacosPanelWindowActivationGuard { app: app.clone() }
 }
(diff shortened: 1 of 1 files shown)


ManthanNimodiya:
@richiemcilroy, got a new approach -> set Accessory once when the first panel window is created, and never restore Regular while panel windows are still visible. 
What do you think is it good or I missed something?

richiemcilroy:
I guess we just need to try it and see if the behaviour of the main window and camera window showing above a fullscreen app is preserved

ManthanNimodiya:
@richiemcilroy,  ok I will try with this approach and let you know at the earliest does that work fine?

ManthanNimodiya:
@richiemcilroy, I ran it and noticed no issues, can you confirm on your side also and lmk for any issues, I applied the suggested approach

richiemcilroy:
hey @greptileai please re-review the PR

richiemcilroy:
Nice work!!

PR #1850, opened 2026-05-21
[codex] cap server proxy upload cap

shashank-sn:
## What changed

- Added a 95 MiB cap to the `/api/upload/signed/proxy` server upload path.
- Return `413` before buffering the request body when `Content-Length` is over the cap.
- Recheck the buffered payload as a fallback when the header is missing or invalid.
- Added the same browser-recorder precheck before using the server proxy fallback.

## Why

The server proxy path buffers media through app/runtime memory before writing to object storage. Large recordings should use the existing direct or multipart object-storage path instead of turning the app server into a media ingest layer.

## Impact

Oversized proxy uploads fail fast with a clear `413` response instead of consuming app runtime memory and risking Cloudflare request/body-limit failures. This keeps video blobs out of SQL/D1-style storage paths and preserves object storage as the media boundary.

## Validation

- `pnpm exec biome check --write 'apps/web/app/api/upload/[...route]/signed.ts' 'apps/web/app/(org)/dashboard/caps/components/web-recorder-dialog/recording-upload.ts'`
- `pnpm --filter @cap/web test -- instant-recording-uploader.test.ts`


shashank-sn:
Closing because this compared the long-lived Cap Video runtime branch against upstream main and showed unrelated history. The scoped fix is being moved to a fork PR against the runtime branch parent so the diff contains only the upload-cap change.

Issue #1773, opened 2026-04-30
Captions toggle in player UI doesn't activate text track on self-hosted


julianwitzel:
**Description**

The Cap video player includes a captions/subtitles UI toggle, but on self-hosted deployments clicking it appears to have no effect. The captions VTT file is correctly:
- Generated by the transcription workflow
- Stored in S3 (`transcription.vtt` next to `result.mp4`)
- Loaded into the DOM as a `<track kind="captions" src="blob:...">` element

However, the track's `mode` stays at `"hidden"` regardless of UI interaction.

**Verified working manually**

Setting the mode via JavaScript makes captions render correctly:
```javascript
document.querySelector('video').textTracks[1].mode = 'showing';
```

This confirms the VTT file, cue parsing, and renderer are all working — only the UI toggle's binding to the track's mode is broken.

**Reproduction**
1. Self-host Cap, record a video with audio
2. Wait for transcription to complete (`transcriptionStatus = 'COMPLETE'` in DB)
3. Open the share page, click the captions toggle in the player
4. Observe: captions don't appear; `track.mode` remains `"hidden"`

**Environment**
- Cap web built from `main` (commit ~52b0acc85 / Apr 2026)
- Custom proxy.ts patches applied for `/embed/` and `/.well-known/` (per #906/PR #1415)
- Custom transcribe.ts patch for #1550
- Browser: Chrome on macOS

Possibly related: Cap uses Vidstack player which has its own captions handling — the toggle button likely sets internal player state but doesn't propagate to `HTMLTrackElement.mode`.

Issue #1356, opened 2025-11-05
Transcription error on self-hosted Cap


bapspatil:
### Description

I've self-hosted Cap and connected my self-hosted Minio as the S3 bucket to store the video files. I added my DEEPGRAM_API_KEY, expecting transcription to work after that, but it's not and I'm currently getting these errors in the server logs:

```
2025-11-05T17:55:12.743690544Z Error transcribing video: Error: Failed to transcribe audio
2025-11-05T17:55:12.743732234Z     at ag (.next/server/chunks/ssr/apps_web_e25a4340._.js:1:32882)
2025-11-05T17:56:24.598063147Z [ShareVideoPage] Starting transcription for video: z0wh037c02gg9be
2025-11-05T17:56:24.778082857Z [transcribeAudio] Starting transcription for URL: https://redacted.domain.com/redacted-coolify-bucket/redacted-id/redacted-subid/result.mp4?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Credential=redacted%2F20251105%2Feu-central%2Fs3%2Faws4_request&X-Amz-Date=20251105T175624Z&X-Amz-Expires=3600&X-Amz-Signature=redacted&X-Amz-SignedHeaders=host&x-amz-checksum-mode=ENABLED&x-id=GetObject
2025-11-05T17:56:25.980731757Z [transcribeAudio] Deepgram transcription error: Error [DeepgramApiError]: {"err_code":"Length Required","err_msg":"Fetching failed to get valid content-length.","request_id":"redacted-2fbf-483d-ba53-redacted"}
2025-11-05T17:56:25.980776927Z     at <unknown> (.next/server/chunks/ssr/apps_web_e25a4340._.js:1:9636) {
2025-11-05T17:56:25.980786547Z   __dgError: true,
2025-11-05T17:56:25.980816747Z   status: 411
2025-11-05T17:56:25.980822467Z }
2025-11-05T17:56:25.981185067Z Error transcribing video: Error: Failed to transcribe audio
2025-11-05T17:56:25.981192267Z     at ag (.next/server/chunks/ssr/apps_web_e25a4340._.js:1:32882)
2025-11-05T17:56:25.981197257Z     at async L (.next/server/chunks/ssr/apps_web_app_s_[videoId]_eae78d05._.js:1:10664)
```

### Reproduction

[...]

PR #1874, opened 2026-05-30
fix(proxy): allow /.well-known/* and /embed/* on self-hosted

### LINKED ISSUE #1768: /embed/* routes redirect to /login on self-hosted, breaking iframe embeds
**Description**

On self-hosted deployments, the `/embed/<videoId>` route incorrectly redirects unauthenticated visitors to `/login`, making iframe embeds unusable. The `/s/<videoId>` share-page route works correctly because it's whitelisted; `/embed/*` was apparently overlooked.

**Reproduction**

1. Self-host Cap with `NEXT_PUBLIC_IS_CAP` unset (or set to anything other than `"true"`)
2. Make a video public
3. Copy the embed code from the share page (e.g. `<iframe src="https://your-cap.example.com/embed/abc123">`)
4. Open the iframe URL directly in a new private/incognito tab — observe the redirect to `/login`

**Expected**: The embed page renders the video player, regardless of authentication state, when the underlying video is public — same as `/s/<videoId>` does today.

**Actual**: Server returns `307 Location: /login`. The iframe loads the login page (or, if the user is logged in, the dashboard, depending on auth state).

**Root cause**

In `apps/web/proxy.ts` (lines 38-56), the self-hosted whitelist allows `/s/`, `/dashboard`, `/api`, `/login`, `/signup`, `/invite`, `/self-hosting`, `/terms`, `/verify-otp` — but **not `/embed/`**:

```javascript
if (buildEnv.NEXT_PUBLIC_IS_CAP !== "true") {
  if (
    !(
      path.startsWith("/s/") ||
      path.startsWith("/middleware") ||
      path.startsWith("/dashboard") ||
      path.startsWith("/onboarding") ||
      path.startsWith("/api") ||
      path.startsWith("/login") ||
      path.startsWith("/signup") ||
      path.startsWith("/invite") ||
      path.startsWith("/self-hosting") ||
      path.startsWith("/terms") ||
      path.startsWith("/verify-otp")
      // /embed missing here
    ) &&
    process.env.NODE_ENV !== "development"
  )
    return NextResponse.redirect(new URL("/login", url.origin));
```

The `/embed/[videoId]/page.tsx` itself uses `provideOptionalAuth` and gracefully handles missing auth, so the page is built correctly — it's just unreachable due to the proxy redirect.

**Fix**

Add `path.startsWith("/embed/")` to the whitelist. One-line change.

**Verified via**

- `curl -i https://my-cap.example.com/embed/<videoId>` returns `307 Location: /login`
- Same behavior with direct IP via `--resolve`, ruling out Cloudflare/CDN
- Same behavior in private browser tabs

**Environment**

- Cap version: `ghcr.io/capsoftware/cap-web:latest` (current as of 2026-04-28)
- Configuration: standard self-hosted Docker compose
- All other features working: Studio Mode recording, share-page playback, auth via Google OAuth + magic link

### LINKED ISSUE #1774: /.well-known/workflow/v1/* routes blocked by proxy whitelist on self-hosted
## Description

The proxy whitelist in `apps/web/proxy.ts` redirects all unwhitelisted paths to `/login` on self-hosted deployments (when `NEXT_PUBLIC_IS_CAP !== "true"`). The `.well-known/workflow/v1/*` routes — used by `@workflow/world-local` for queue dispatch and step execution — aren't in the whitelist.

When `start(transcribeVideoWorkflow, ...)` (or any other workflow `start()` call) tries to POST to `/.well-known/workflow/v1/flow`, it gets a `307 → /login` instead of the expected workflow response. The dispatch fails silently, the workflow never executes, and any feature relying on workflows (transcription, AI summaries) stalls.

## Reproduction

1. Self-host Cap with `NEXT_PUBLIC_IS_CAP` unset
2. From inside the cap-web container:
```bash
   wget -qO- http://127.0.0.1:3000/.well-known/workflow/v1/step
```
   Returns the login page HTML instead of a workflow response.
3. Or trigger a transcription: it stalls at `transcriptionStatus = "PROCESSING"` indefinitely, no `/audio/extract` calls reach the media-server.

## Root cause

In `apps/web/proxy.ts` lines ~40-56:

```js
if (buildEnv.NEXT_PUBLIC_IS_CAP !== "true") {
  if (
    !(
      path.startsWith("/s/") ||
      path.startsWith("/middleware") ||
      path.startsWith("/dashboard") ||
      path.startsWith("/onboarding") ||
      path.startsWith("/api") ||
      path.startsWith("/login") ||
      path.startsWith("/signup") ||
      path.startsWith("/invite") ||
      path.startsWith("/self-hosting") ||
      path.startsWith("/terms") ||
      path.startsWith("/verify-otp")
      // /.well-known/ missing — blocks workflow internal routes
      // /embed/ also missing — see #906 / PR #1415
    ) &&
    process.env.NODE_ENV !== "development"
  )
    return NextResponse.redirect(new URL("/login", url.origin));
```

## Fix

Same shape as #906:

```diff
       path.startsWith("/verify-otp") ||
+      path.startsWith("/embed/") ||
+      path.startsWith("/.well-known/")
```

The `/.well-known/` path is reserved by RFC 8615 for well-known URIs and is a sensible whitelist entry generally — workflow routes, ATproto verification, etc. all live under it.

## Verified

After applying both whitelist additions plus the workflow-race fix from #1550, the full transcription + AI pipeline runs end-to-end on self-hosted (Hetzner CPX41, `cap-web` built from current `main`).

## Environment

- Cap version: built from `main` commit 52b0acc85
- Self-hosted via Docker compose (Hetzner)
- Other services in stack: Hetzner S3, MySQL 8, cap-media-server (latest)

kovashikawa:
## Summary

- Two-line fix: add `/.well-known/` and `/embed/` to the self-host proxy whitelist.
- Restores all workflow execution on self-hosted Cap (transcription, AI generation, etc.) and fixes broken iframe embeds.

## Root cause

`apps/web/proxy.ts:40-58` redirects every path not in an allow-list to `/login` when `NEXT_PUBLIC_IS_CAP !== "true"` (i.e., self-hosted builds). `/.well-known/` and `/embed/` were both missing from that list.

### `/.well-known/` (workflows broken)

The Vercel Workflows runtime that Cap uses (`apps/web/app/.well-known/workflow/v1/{step,flow,webhook}/route.js`) dispatches workflow steps by issuing HTTP requests back to itself at those routes. On self-host, those requests get a 307 → `/login`, so workflows are enqueued but never run.

Visible symptom: `transcribeVideo()` is called and logs `[transcribeVideo] Triggering transcription workflow…`, but no subsequent `[transcribe]` step lines ever appear, no Deepgram/Whisper request is made, and `videos.transcriptionStatus` stays `NULL` forever. Same for `aiGenerationStatus`.

### `/embed/` (iframe embeds broken)

`<iframe src=".../embed/...">` embeds on self-hosted instances load the login page instead of the embedded video, because the embed route also fails the proxy whitelist check.

## Fix

```ts
// apps/web/proxy.ts
path.startsWith("/verify-otp") ||
path.startsWith("/.well-known/") ||  // ← added (commit 1)
path.startsWith("/embed/")           // ← added (commit 2)
```

That's it. The matcher in `config.matcher` already excludes `_next/*` etc., and both prefixes are scoped to existing app routes, so widening is safe.

## Verification

Reproduced on a local Docker Compose self-host (`docker-compose.yml`).

**Before this patch**
- `curl http://localhost:3000/.well-known/workflow/v1/manifest.json` returns the Next.js HTML shell (the redirect-to-login response).
- Uploading a video, then triggering transcription via the share page or `POST /api/videos/:id/retry-transcription`, leaves `transcriptionStatus = NULL` forever.

**After this patch**
- Same `curl` returns the workflow manifest JSON (correct route).
- Share-page render fires `transcribeVideo()`; the workflow runs end-to-end: audio extraction → transcription provider → `transcription.vtt` written to S3 → `transcriptionStatus = COMPLETE`. AI generation then runs and writes `metadata.summary` + `metadata.chapters`.

## Related

- Fixes #1774 (`/.well-known/`)
- Fixes #1768 (`/embed/`)
- Root cause behind #1356 and #1550 (self-host transcription/AI never runs)
- May transitively address #1773 (self-host captions toggle), since captions depend on the transcription pipeline running. Not independently verified.

<h3>Greptile Summary</h3>

Two-line whitelist addition to the self-hosted proxy gate in `proxy.ts`, restoring `/.well-known/` workflow callbacks and `/embed/` iframe routes that were incorrectly redirected to `/login`.

- `/.well-known/` allows the Vercel Workflows runtime to call back to its own step/flow/webhook routes without being bounced; without this, transcription and AI generation workflows are silently enqueued but never execute.
- `/embed/` allows unauthenticated access to embedded video player routes, which is the correct behaviour for public embeds on self-hosted instances.

Safe to merge; the fix is minimal, targeted, and correctly scoped to routes that must be publicly reachable on self-hosted deployments.

The change is two lines in one file and the logic is easy to verify against the described failure. The only open question is whether the `/.well-known/workflow/v1/webhook` handler validates Vercel's request signature independently, since prefix matching now makes the entire `/.well-known/` tree reachable without the proxy auth check.

The workflow route handlers under `apps/web/app/.well-known/workflow/v1/` are worth a quick glance to confirm they verify the incoming request signature before acting.

| Filename | Overview |
|----------|----------|
| apps/web/proxy.ts | Adds `/.well-known/` and `/embed/` to the self-host proxy auth bypass list; straightforward fix with prefix-match breadth worth noting |

</details>

`````markdown

---

### Issue 1 of 1
apps/web/proxy.ts:54-55
**Broad bypass scope for future routes**

Both new entries use prefix matching, so every future route added under `/.well-known/` (including any webhook handler) and `/embed/` will automatically bypass the self-host auth redirect — with no indication at the call site that this is happening. The `/.well-known/workflow/v1/webhook` route in particular is publicly reachable as a result; its safety depends entirely on Vercel Workflow's own request-signing, which isn't visible here. Worth confirming the webhook handler validates the signature before relying on this being safe.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(proxy): allow /embed/\* on self-hoste..."](https://github.com/capsoftware/cap/commit/5a3ea5c4b96d3f95c4288bd9407747474689f694) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34750272)</sub>

### DIFF
--- a/apps/web/proxy.ts
+++ b/apps/web/proxy.ts
@@ -51,7 +51,8 @@ export async function proxy(request: NextRequest) {
 				path.startsWith("/download") ||
 				path.startsWith("/terms") ||
 				path.startsWith("/verify-otp") ||
-				path.startsWith("/.well-known/")
+				path.startsWith("/.well-known/") ||
+				path.startsWith("/embed/")
 			) &&
 			process.env.NODE_ENV !== "development"
 		)
--- a/apps/web/proxy.ts
+++ b/apps/web/proxy.ts
@@ -51,7 +51,7 @@ export async function proxy(request: NextRequest) {
 				path.startsWith("/download") ||
 				path.startsWith("/terms") ||
 				path.startsWith("/verify-otp") ||
-				path.startsWith("/.well-known/") ||
+				path.startsWith("/.well-known/workflow/") ||
 				path.startsWith("/embed/")
 			) &&
 			process.env.NODE_ENV !== "development"
--- a/apps/web/proxy.ts
+++ b/apps/web/proxy.ts
@@ -50,7 +50,8 @@ export async function proxy(request: NextRequest) {
 				path.startsWith("/self-hosting") ||
 				path.startsWith("/download") ||
 				path.startsWith("/terms") ||
-				path.startsWith("/verify-otp")
+				path.startsWith("/verify-otp") ||
+				path.startsWith("/.well-known/")
 			) &&
 			process.env.NODE_ENV !== "development"
 		)


PicchiKevin:
nice!

Erudition:
also Add /rive/ and /sounds/ to the path allowlist

PR #1884, opened 2026-06-04
fix(web): exclude .well-known from proxy matcher so self-hosted workflows run

mstublefield:
## Summary

On self-hosted builds (`NEXT_PUBLIC_IS_CAP` unset), `apps/web/proxy.ts` redirects every path outside its allowlist to `/login`. The workflow runtime's local queue self-invokes via `POST /.well-known/workflow/v1/flow` and `/step` — those requests get 307'd to the login page, so every `start()`-ed workflow run sits in `pending` forever and uploaded videos never process. This affects the published `ghcr.io/capsoftware/cap-web:latest` image (reproduced there directly).

Nobody noticed before because:
- **cap.so production**: the `IS_CAP === "true"` branch passes everything through
- **local dev**: the redirect is skipped when `NODE_ENV !== "production"`

Only self-hosted Docker deployments hit it.

## Fix

One line — exclude `.well-known` in the proxy matcher, alongside the existing `api` exclusion. Machine-to-machine endpoints (workflow queue, workflow webhooks) shouldn't pass through page-auth or custom-domain logic at all.

## Repro / verification (local self-host stack: cap-web + media-server + MySQL + MinIO)

Before:
```
$ curl -s -X POST -D - -o /dev/null http://localhost:3001/.well-known/workflow/v1/flow | grep -E "HTTP|location"
HTTP/1.1 307 Temporary Redirect
location: /login
```
Uploaded video → workflow run stuck `"status": "pending"` in `.workflow-data/runs/`, `video_uploads` row frozen at "Starting video processing...".

After:
```
HTTP/1.1 400 Bad Request
{"error":"Missing required headers"}   ← reaches the workflow handler
```
Fresh upload → `processVideoWorkflow` run reaches `"status": "completed"`, media-server processes the video, result/thumbnail/preview artifacts land in storage, webhook populates duration/dimensions.

<h3>Greptile Summary</h3>

This one-line fix excludes `/.well-known` paths from the Next.js middleware matcher, preventing the proxy from redirecting workflow self-invocation requests (`POST /.well-known/workflow/v1/flow` and `/step`) to `/login` on self-hosted deployments where `NEXT_PUBLIC_IS_CAP` is unset.

- Adds `|.well-known` to the existing negative lookahead in the `config.matcher`, mirroring the `api` exclusion already present, so machine-to-machine workflow endpoints bypass page-auth and custom-domain logic entirely.
- The fix is correctly scoped: cap.so production is unaffected (it takes the `IS_CAP === "true"` fast path), local dev is unaffected (redirect is skipped outside production), and only Docker self-hosted deployments were broken.

Safe to merge — changes only the middleware matcher, has no runtime logic side-effects, and the fix has been manually verified against the broken self-hosted workflow path.

The change is minimal and well-targeted. The dot in `.well-known` is unescaped in the regex lookahead (as are the dots in `favicon.ico`, `robots.txt`, and `sitemap.xml`), which works in practice but is technically imprecise. For machine-to-machine endpoints this is worth tightening, though it doesn't affect the correctness of the fix for the intended path.

Only `apps/web/proxy.ts` changed; the regex lookahead in the matcher line is the one spot that warrants a second look.

| Filename | Overview |
|----------|----------|
| apps/web/proxy.ts | Adds `.well-known` to the Next.js middleware matcher exclusion list so workflow self-invocation requests are no longer intercepted and redirected to /login on self-hosted deployments. |

</details>

`````markdown

---

### Issue 1 of 1
apps/web/proxy.ts:118
In a regex negative lookahead, `.` is a wildcard that matches any character, not a literal dot. `(?!.well-known)` therefore excludes any path whose second character is followed by `well-known` (e.g., `/awell-known`), not strictly `/.well-known`. It works for the intended case but is inconsistent with intent. The same issue exists for `favicon.ico`, `robots.txt`, and `sitemap.xml`, but the self-hosted workflow paths make this entry more security-relevant, so it's worth being explicit.

```suggestion
		"/((?!api|_next/static|_next/image|favicon\\.ico|robots\\.txt|sitemap\\.xml|\\.well-known).*)",
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): exclude .well-known from proxy..."](https://github.com/capsoftware/cap/commit/4933ca7421da74b020781de3cd864b2747621828) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35640367)</sub>


richiemcilroy:
hey @greptileai please re-review

PR #1890, opened 2026-06-06
feat(folders): public sharing of a folder via a signed link

alexis-morain:
## Summary

Adds a per-folder **Share folder** action that produces a public URL anyone with the link can open. The visitor sees a clean index of the folder's videos (thumbnails, names, durations, dates) and clicks through to the existing `/s/<videoId>` player — no account, no signup.

I built this for a self-hosted instance where I needed to hand a client all the videos in one project folder without giving them a Cap account. Felt generally useful so I'm upstreaming it.

## Architecture

- **Schema** — single new column `folders.publicShared boolean default false`. No new table; the slug is derived, not stored.
- **Slug** — `base64url(folderId) + \".\" + hmac(folderId, NEXTAUTH_SECRET)`. Stateless. Rotating `NEXTAUTH_SECRET` revokes every outstanding folder-share link.
- **Action** — `enableFolderSharing(folderId)` flips `folders.publicShared = true` AND `videos.public = true` for every video in the folder so the existing share page keeps working unauthenticated. `disableFolderSharing` reverts both.
- **Public route** — `/share/f/[slug]/page.tsx` (Server Component): verifies the slug, refuses if `publicShared = false`, resolves S3 signed thumbnail URLs server-side (the existing `/api/thumbnail` returns JSON, not the image), renders a responsive grid linking each card to `/s/<videoId>`.
- **Proxy** — `/share/` added to the self-hosted whitelist next to `/embed/` so the route is reachable without auth.
- **UI** — new \"Share folder\" item in the existing `FoldersDropdown` kebab menu (hidden for folders owned by a Space — out of scope for this PR). New `ShareFolderDialog` with state hydration, Enable / Disable buttons, copy-to-clipboard.

## Test plan

- [ ] On `/dashboard/caps`, kebab on any folder → **Share folder** → dialog opens, shows \"Enable sharing\".
- [ ] Click **Enable sharing** → toast confirms, public link appears in the dialog and is copied to clipboard.
- [ ] Open the link in an incognito window → folder index renders with thumbnails / dates / durations, no auth prompt.
- [ ] Click a video → `/s/<videoId>` plays without login.
- [ ] Re-open the dialog on the original folder → still shows the same link.
- [ ] Click **Disable sharing** → toast confirms, opening the public link returns 404; opening any `/s/<videoId>` from that folder requires auth again (videos flipped back to private).
- [ ] Repeat after rotating `NEXTAUTH_SECRET` → previously-generated links return 404. Generating a new link produces a different slug.

## Notes for reviewers

- **Migration not regenerated.** No MySQL on the dev machine that pushed this. Please run `pnpm db:generate` to emit the migration file and journal entry; the schema change is just the new boolean column.
- **\"Set every video to public\" trade-off.** This is the simplest path to making `/s/<videoId>` work for guests without touching the share page. Downside: someone who guesses a video id while a folder is shared can hit it directly. A follow-up could swap this for a signed query param `/s/<videoId>?fs=<slug>` checked against the folder slug, keeping the videos technically private. Kept out of this PR to limit the blast radius.
- **Spaces folders.** Skipped intentionally — the kebab passes `setShareDialogOpen={spaceId ? undefined : ...}`. The Space sharing model is a separate question and I didn't want to overload it here.
- **No password / no expiry.** Conscious v1 scope. Both are clean additions on top (extra column for password hash, extra column for expiresAt) without architecture changes.

<h3>Greptile Summary</h3>

This PR adds stateless public folder sharing: a signed slug (`base64url(folderId) + \".\" + truncated-HMAC`) is generated from `NEXTAUTH_SECRET`, and a new `/share/f/[slug]` page renders the folder's video grid for unauthenticated visitors.

- **`disableFolderSharing` is destructive**: it unconditionally bulk-sets all videos in the folder to `public = false`, which permanently makes private any video that was already individually shared before folder sharing was ever enabled. Only videos whose `public` flag was flipped *by* `enableFolderSharing` should be reverted.
- **Timing-unsafe HMAC verification**: `verifyFolderShareSlug` uses `!==` to compare the expected and received HMAC signatures; `crypto.timingSafeEqual` should be used instead.
- The thumbnail resolver in the share page issues 2 sequential S3 API calls (`listObjects` + `getSignedObjectUrl`) for every video; for large folders this causes meaningful latency and could be reduced by constructing the key directly.

Not safe to merge as-is: disabling folder sharing will silently revoke individually-shared videos, and the HMAC verification is not constant-time.

Two issues need resolution before merge. The `disableFolderSharing` action bulk-sets every video in the folder to private without checking whether those videos were already public before folder sharing was enabled — any video a user had independently shared will silently lose its public access. The HMAC verification in `folder-share.ts` uses a plain string comparison instead of `timingSafeEqual`, introducing a timing oracle on a security-critical path.

`apps/web/actions/folders/share-folder.ts` (destructive video visibility reset) and `apps/web/lib/folder-share.ts` (non-constant-time signature check) both need fixes before this lands.

<details open><summary><h3>Security Review</h3></summary>

- **Timing-unsafe HMAC comparison** (`apps/web/lib/folder-share.ts` line 37): `sign(folderId) !== sig` uses a short-circuit string comparison, leaking byte-position timing information. Because the HMAC check runs before any database query, an attacker can probe it freely without needing a real folder ID. Replacing it with `crypto.timingSafeEqual` eliminates the timing oracle.
</details>

| Filename | Overview |
|----------|----------|
| apps/web/lib/folder-share.ts | New HMAC-based slug signing/verification module; uses non-constant-time string comparison in `verifyFolderShareSlug`, which should use `timingSafeEqual` instead. |
| apps/web/actions/folders/share-folder.ts | Server actions for enabling/disabling folder sharing; `disableFolderSharing` unconditionally sets all folder videos to private, destroying the pre-existing public state of videos that were already shared individually before folder sharing was enabled. |
| apps/web/app/share/f/[slug]/page.tsx | New public folder share page; verifies slug and renders video grid, but issues 2 S3 API calls per video for thumbnail resolution and uses `as any` casts to work around Drizzle branded-type mismatches. |
| apps/web/app/(org)/dashboard/caps/components/ShareFolderDialog.tsx | New client dialog for managing folder sharing state; hydration pattern and clipboard handling look correct; minor: clipboard failure is silently swallowed in `handleEnable`. |
| apps/web/app/(org)/dashboard/caps/components/FoldersDropdown.tsx | Adds optional "Share folder" menu item; correctly hidden for space-owned folders via `undefined` prop guard. |
| packages/database/schema.ts | Adds `publicShared boolean NOT NULL DEFAULT false` column to `folders` table; schema change is straightforward, migration not yet generated (noted in PR). |
| apps/web/proxy.ts | Adds `/share/` to the self-hosted auth bypass whitelist alongside `/embed/`; minimal and correct change. |
| apps/web/app/(org)/dashboard/caps/components/Folder.tsx | Wires `ShareFolderDialog` into the folder card; correctly gates the share action behind `!spaceId`. |

</details>

`````markdown
Fix the following 5 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 5
apps/web/actions/folders/share-folder.ts:59-62
**`disableFolderSharing` permanently silences pre-existing public videos**

`disableFolderSharing` unconditionally sets `public = false` on every video in the folder, regardless of whether those videos were already public before folder sharing was enabled. A user who individually shared a video via `/s/<videoId>` before ever touching the folder-share feature will silently find that video made private the moment they disable folder sharing — with no way to know it happened.

`enableFolderSharing` should record which videos it flipped (e.g., by filtering on `public = false` before updating, or by storing a flag like `sharedViaFolder`), so `disableFolderSharing` can revert only those videos rather than all of them.

### Issue 2 of 5
apps/web/lib/folder-share.ts:1
The `timingSafeEqual` import is missing when switching to constant-time comparison.

```suggestion
import { createHmac, timingSafeEqual } from "node:crypto";
```

### Issue 3 of 5
apps/web/lib/folder-share.ts:37
**Non-constant-time HMAC verification is susceptible to timing attacks**

`sign(folderId) !== sig` is a standard string comparison that short-circuits on the first differing byte, leaking timing information. Since the HMAC check runs before any DB lookup, an attacker can probe it freely. The fix is to use `timingSafeEqual` from `node:crypto`, which always runs in constant time regardless of where the strings differ.

```suggestion
	const expected = Buffer.from(sign(folderId));
	const actual = Buffer.from(sig);
	if (expected.length !== actual.length || !timingSafeEqual(expected, actual))
		return null;
```

### Issue 4 of 5
apps/web/app/share/f/[slug]/page.tsx:47-71
**N+1 S3 API calls on page load**

`resolveThumbnailUrl` makes two sequential S3 API calls per video (`listObjects` then `getSignedObjectUrl`). For a folder with 30 videos, the page issues 60 S3 calls concurrently via `Promise.all` — each of which fans out to the cloud. This will cause noticeably slow page loads and may hit S3 rate limits on large folders. If the thumbnail key pattern is stable (e.g., `{ownerId}/{videoId}/screen-capture.jpg`), the `listObjects` call can be skipped and the signed URL generated directly from the known key.

### Issue 5 of 5
apps/web/app/share/f/[slug]/page.tsx:24
**`as any` casts suppress Drizzle type safety**

Both WHERE clauses cast `folderId` to `any` (lines 24, 88, 105). This suggests a branded-type mismatch between the decoded string returned by `verifyFolderShareSlug` and the column's declared type (`Folder.FolderId`). The right fix is to cast the return value of `verifyFolderShareSlug` to the correct branded type once, or to type the function return as `Folder.FolderId | null`, instead of silencing the type checker at every call site.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(folders): public sharing of a folde..."](https://github.com/capsoftware/cap/commit/45a652dee5c1e93fbdb2226f622169d6ae7d4087) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35998148)</sub>

> Greptile also left **5 inline comments** on this PR.

### DIFF
--- a/apps/web/actions/folders/share-folder.ts
+++ b/apps/web/actions/folders/share-folder.ts
@@ -43,8 +43,10 @@ export async function enableFolderSharing(folderId: Folder.FolderId) {
 		.where(eq(folders.id, folderId));
 	await db()
 		.update(videos)
-		.set({ public: true })
-		.where(eq(videos.folderId, folderId));
+		.set({ public: true, publicSourcedFromFolderShare: true })
+		.where(
+			and(eq(videos.folderId, folderId), eq(videos.public, false)),
+		);
 	revalidatePath(`/dashboard/folder/${folderId}`);
 	revalidatePath(`/dashboard/caps`);
 	return { slug: signFolderShareSlug(folderId) };
@@ -58,8 +60,13 @@ export async function disableFolderSharing(folderId: Folder.FolderId) {
 		.where(eq(folders.id, folderId));
 	await db()
 		.update(videos)
-		.set({ public: false })
-		.where(eq(videos.folderId, folderId));
+		.set({ public: false, publicSourcedFromFolderShare: false })
+		.where(
+			and(
+				eq(videos.folderId, folderId),
+				eq(videos.publicSourcedFromFolderShare, true),
+			),
+		);
 	revalidatePath(`/dashboard/folder/${folderId}`);
 	revalidatePath(`/dashboard/caps`);
 }
--- a/apps/web/actions/folders/share-folder.ts
+++ b/apps/web/actions/folders/share-folder.ts
@@ -1,0 +1,65 @@
+"use server";
+
+import { db } from "@cap/database";
+import { getCurrentUser } from "@cap/database/auth/session";
+import { folders, videos } from "@cap/database/schema";
+import type { Folder } from "@cap/web-domain";
+import { and, eq } from "drizzle-orm";
+import { revalidatePath } from "next/cache";
+import { signFolderShareSlug } from "@/lib/folder-share";
+
+const requireOwnedFolder = async (folderId: Folder.FolderId) => {
+	const user = await getCurrentUser();
+	if (!user || !user.activeOrganizationId)
+		throw new Error("Unauthorized or no active organization");
+
+	const [folder] = await db()
+		.select()
+		.from(folders)
+		.where(
+			and(
+				eq(folders.id, folderId),
+				eq(folders.organizationId, user.activeOrganizationId),
+			),
+		);
+
+	if (!folder) throw new Error("Folder not found");
+	return folder;
+};
+
+export async function getFolderShareState(folderId: Folder.FolderId) {
+	const folder = await requireOwnedFolder(folderId);
+	return {
+		isShared: folder.publicShared === true,
+		slug: folder.publicShared ? signFolderShareSlug(folderId) : null,
+	};
+}
+
+export async function enableFolderSharing(folderId: Folder.FolderId) {
+	await requireOwnedFolder(folderId);
+	await db()
+		.update(folders)
+		.set({ publicShared: true })
+		.where(eq(folders.id, folderId));
+	await db()
+		.update(videos)
+		.set({ public: true })
+		.where(eq(videos.folderId, folderId));
+	revalidatePath(`/dashboard/folder/${folderId}`);
+	revalidatePath(`/dashboard/caps`);
+	return { slug: signFolderShareSlug(folderId) };
+}
+
+export async function disableFolderSharing(folderId: Folder.FolderId) {
+	await requireOwnedFolder(folderId);
+	await db()
+		.update(folders)
+		.set({ publicShared: false })
+		.where(eq(folders.id, folderId));
+	await db()
+		.update(videos)
+		.set({ public: false })
+		.where(eq(videos.folderId, folderId));
+	revalidatePath(`/dashboard/folder/${folderId}`);
+	revalidatePath(`/dashboard/caps`);
+}
--- a/apps/web/app/(org)/dashboard/caps/components/Folder.tsx
+++ b/apps/web/app/(org)/dashboard/caps/components/Folder.tsx
@@ -14,6 +14,7 @@ import { ConfirmationDialog } from "../../_components/ConfirmationDialog";
 import { useDashboardContext, useTheme } from "../../Contexts";
 import { registerDropTarget } from "../../folder/[id]/components/ClientCapCard";
 import { FoldersDropdown } from "./FoldersDropdown";
+import { ShareFolderDialog } from "./ShareFolderDialog";
 
 export type FolderDataType = {
 	name: string;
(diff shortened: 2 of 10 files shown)


alexis-morain:
Closing this in favor of #1906 ("Public links for spaces and folders"), which was merged on 2026-06-10 and covers the same capability more completely: public collection pages for both folders and spaces at `/c/[id]`, with optional password protection, email-domain restriction, custom branding and a configurable layout. This PR (folder-only public sharing via a signed `/share/f/[slug]` link) is now redundant. Thanks!

PR #1893, opened 2026-06-06
Parallelize startup and harden playback/export A/V sync

richiemcilroy:
- Parallelize editor startup (mic/system-audio load + GPU layer warmup overlap decoder creation)
  - Replace audio-timing-repair log-scraping with structured AudioGapSummary metadata (recorder → AudioMeta → editor)
  - Harden audio mixer against offset overflow (i128 intermediates) and skip decode work for sub-fallback frames
  - Synthesize a default timeline for un-edited recordings (decoded media duration) so raw/CLI exports render
  - Gapless export audio on zero-frame fallback; fragmented-recording remux on finalize/export
  - Tighten the audio-ready handshake timeout to prevent unsynced playback

<h3>Greptile Summary</h3>

This PR hardens A/V sync across playback and export through several coordinated changes: structured `AudioGapSummary` metadata replaces log-scraping for startup-offset repair, GPU layer warmup is parallelised with decoder creation, the video clock is now stamped only after the audio device fires its first callback, and a default timeline is synthesised from encoded frame counts at recording stop so un-edited recordings are immediately exportable.

- **A/V sync**: `start = Instant::now()` moved after `AudioPlayback::spawn()` returns, which blocks on the first device callback via a new channel handshake; initial latency offset switched from `initial_compensation_secs` (2× safety multiplier) to raw `initial_output_latency_secs`; silent audio frame substituted for `None` renders during export to prevent gaps.
- **`AudioGapSummary` pipeline**: structured `startup_overlap_trimmed_ms` / `startup_overlap_drops` counters are accumulated in `AudioGapTracker`, stored in a `OnceLock` on `FinishedOutputPipeline`, written into `AudioMeta.gap_summary`, and consumed by `audio_timing_repair_offset` in the editor—replacing fragile log-scraping while keeping a legacy fallback for old recordings.
- **Overflow hardening**: audio mixer uses `i128` intermediates and `usize::try_from` instead of `isize` and `wrapping_add_signed` to prevent index overflow on large negative track offsets.

Safe to merge; the changes are well-tested and correctness-focused, with one open concern around startup latency for very long recordings.

The audio timing-repair rewrite is thorough and well-covered by new unit tests. The `start = Instant::now()` move correctly ties video to the audio device's first callback, fixing the sync regression. The one remaining concern is that `AudioPlayback::spawn()` has no upper-bound timeout for the `PrerenderedAudioBuffer::new()` pre-render phase: for a multi-hour recording the polling loop can silently stall for tens of seconds before playback begins.

crates/editor/src/playback.rs — the spawn() polling loop should have an outer timeout covering the pre-render phase, not just AUDIO_FIRST_CALLBACK_TIMEOUT which starts after stream.play().

| Filename | Overview |
|----------|----------|
| crates/editor/src/playback.rs | Moves `start = Instant::now()` after `audio_playback.spawn()` to synchronize A/V start; adds first-callback handshake with `AUDIO_FIRST_CALLBACK_TIMEOUT`; hardens `prefetch_duration` and `duration` against invalid timelines; changes initial latency from `initial_compensation_secs` (2× multiplier) to raw `initial_output_latency_secs`. |
| crates/editor/src/editor_instance.rs | Replaces log-scraping A/V timing repair with structured `AudioGapSummary` metadata; adds `LegacyAudioTimingRepair` fallback for old recordings; parallelises audio I/O via `spawn_audio_load`; removes 5 s hardcoded fallback duration; minor redundant guard in `audio_timing_repair_offset`. |
| crates/recording/src/output_pipeline/core.rs | Adds `AudioGapSummary` struct and per-event startup/total accounting to `AudioGapTracker`; passes `frame_count` to `record_overlap`; stores summary in `OnceLock` on `OutputPipeline` and surfaces it in `FinishedOutputPipeline`. |
| crates/recording/src/studio_recording.rs | Synthesizes `TimelineConfiguration` and `ClipConfiguration` from encoded frame_count/fps at recording stop; propagates `gap_summary` into `AudioMeta`; persists populated project config so un-edited recordings have a valid timeline for export/playback. |
| crates/audio/src/renderer.rs | Replaces i32/isize arithmetic with i128 intermediates and `usize::try_from` to prevent overflow on large offsets; fixes stereo base-index calculation; adds regression test for negative offset (leading-silence) behaviour. |
| crates/export/src/mp4.rs | Emits a silent audio frame instead of skipping on zero-frame fallback, ensuring gapless export when the audio renderer returns None for a given time slot. |
| crates/editor/src/editor.rs | Splits `start_renderer_layers_creation` into start + `finish_renderer_layers_creation` so GPU layer warmup overlaps decoder/segment creation; removes now-unused `total_frames` field. |
| crates/recording/src/feeds/microphone.rs | Adds `normalize_pending_timestamps` to correct stale startup burst timestamps before they are enqueued, preventing phantom overlap events that would otherwise inflate the gap summary. |
| crates/rendering/src/decoder/avassetreader.rs | Computes `minimum_fallback_frame` from pending requests and skips decode work for frames below that bound, reducing unnecessary decode overhead during seek. |

</details>

1. `crates/editor/src/editor_instance.rs`, line 176-180 ([link](https://github.com/capsoftware/cap/blob/3d42742d6057a2c870fa717fd3d58a053b2338b0/crates/editor/src/editor_instance.rs#L176-L180)) 

   <a href="#">P2</a> **Hardcoded 5.0 s fallback creates incorrect timeline for any recording not ~5 s long**

   When both `Video::new` and `get_video_duration_fallback` fail (e.g., partially written or corrupt display file), the synthesized `TimelineSegment` uses `end: 5.0`. A 30-second recording whose display file is unreadable would export with a 5-second timeline, silently truncating content. Since this fallback is exercised for real on first open (before recovery or remux), the project config is then written with the wrong `end` value and persisted. Returning `None` from the filter in the `MultipleSegments` arm (same as the explicit `duration <= 0.0` guard just below) and warning the user would be safer than silently synthesising an arbitrary duration.

   `````markdown
   Path: crates/editor/src/editor_instance.rs
   Line: 176-180

   Comment:
   **Hardcoded 5.0 s fallback creates incorrect timeline for any recording not ~5 s long**

   When both `Video::new` and `get_video_duration_fallback` fail (e.g., partially written or corrupt display file), the synthesized `TimelineSegment` uses `end: 5.0`. A 30-second recording whose display file is unreadable would export with a 5-second timeline, silently truncating content. Since this fallback is exercised for real on first open (before recovery or remux), the project config is then written with the wrong `end` value and persisted. Returning `None` from the filter in the `MultipleSegments` arm (same as the explicit `duration <= 0.0` guard just below) and warning the user would be safer than silently synthesising an arbitrary duration.

   `````
   </details>
</details>

`````markdown

---

### Issue 1 of 2
crates/editor/src/playback.rs:1285-1356
**Pre-render phase has no upper-bound timeout in `spawn()`**

`AudioPlayback::spawn()` polls `ready_rx` every 50 ms and only exits when the audio thread sends a value or `stop_rx` fires. The audio thread sends `ready_tx` only _after_ `PrerenderedAudioBuffer::new()` completes **and** the first device callback arrives. `PrerenderedAudioBuffer::new()` is O(recording length), so for a multi-hour recording on modest hardware it can take well over 10–20 s. During that entire window the polling loop spins with no way out except user cancellation.

`AUDIO_FIRST_CALLBACK_TIMEOUT` does not help here: it only starts timing after `stream.play()`, which is called _after_ pre-render finishes. The net effect is that pressing play on a long recording can silently stall for tens of seconds before video begins—trading the previous async desync for a synchronous startup freeze.

### Issue 2 of 2
crates/editor/src/editor_instance.rs:759-762
The `overlap_dropped_frames < MIN_STALE_STARTUP_DROPS` guard is unreachable: `startup_overlap_drops ≤ overlap_dropped_frames` by construction, so if the `startup_overlap_drops` check passes (`>= 3`), `overlap_dropped_frames >= 3` is guaranteed. Removing the redundant condition avoids misleading future readers into thinking total drops could prevent repair when startup drops are sufficient.

```suggestion
    if summary.startup_overlap_drops < MIN_STALE_STARTUP_DROPS
        || !(MIN_STALE_STARTUP_TRIMMED_MS..=MAX_STALE_STARTUP_REPAIR_MS)
            .contains(&summary.startup_overlap_trimmed_ms)
```


`````

</details>

<sub>Reviews (3): Last reviewed commit: ["Add test for negative timing offset cons..."](https://github.com/capsoftware/cap/commit/105ddce09a3f86569acf63158c9ffa26f3478bc0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36040067)</sub>

### DIFF
--- a/apps/desktop/src-tauri/examples/desktop-display-transport-benchmark.rs
+++ b/apps/desktop/src-tauri/examples/desktop-display-transport-benchmark.rs
@@ -8,7 +8,8 @@ use std::{
 use cap_desktop_lib::frame_ws::{WSFrame, WSFrameFormat, create_watch_frame_ws};
 use cap_editor::{
     EditorFrameOutput, Playback, PlaybackRenderOutputFormat, PlaybackSkipReason, PlaybackTelemetry,
-    PlaybackTelemetryEvent, Renderer, start_renderer_layers_creation,
+    PlaybackTelemetryEvent, Renderer, finish_renderer_layers_creation,
+    start_renderer_layers_creation,
 };
 use cap_project::{
     ProjectConfiguration, RecordingMeta, RecordingMetaInner, StudioRecordingMeta,
@@ -229,7 +230,7 @@ async fn main() {
 
     tokio::time::sleep(Duration::from_millis(startup_delay_ms)).await;
 
-    let layers_rx = start_renderer_layers_creation(&render_constants);
+    let layers_rx = start_renderer_layers_creation(&render_constants, &project);
     let segment_medias =
         match cap_editor::create_segments(&recording_meta, meta.as_ref(), false).await {
             Ok(segments) => Arc::new(segments),
@@ -238,6 +239,7 @@ async fn main() {
                 std::process::exit(1);
             }
         };
+    let layers_rx = finish_renderer_layers_creation(layers_rx).await;
 
     let (telemetry, mut telemetry_rx) = PlaybackTelemetry::channel();
     let (frame_tx, mut frame_rx) = mpsc::unbounded_channel::<usize>();
--- a/apps/desktop/src-tauri/src/export.rs
+++ b/apps/desktop/src-tauri/src/export.rs
@@ -715,13 +715,19 @@ fn resolve_exporter_binary() -> Result<PathBuf, String> {
 }
 
 fn exporter_binary_candidates(root: &Path) -> Vec<PathBuf> {
-    let mut candidates = vec![
+    let mut candidates = Vec::new();
+
+    if cfg!(debug_assertions) {
+        candidates.extend(debug_exporter_binary_candidates(root));
+    }
+
+    candidates.push(
         root.join("apps")
             .join("desktop")
             .join("src-tauri")
             .join("binaries")
             .join(exporter_bin_name()),
-    ];
+    );
 
     if let Some(target_triple) = current_target_triple() {
         candidates.push(
@@ -734,6 +740,36 @@ fn exporter_binary_candidates(root: &Path) -> Vec<PathBuf> {
                     std::env::consts::EXE_SUFFIX
                 )),
         );
+    }
+
+    candidates
+}
+
+fn debug_exporter_binary_candidates(root: &Path) -> Vec<PathBuf> {
+    let mut candidates = vec![
+        root.join("target").join("debug").join(exporter_bin_name()),
+        root.join("target")
+            .join("debug")
+            .join(format!("cap{}", std::env::consts::EXE_SUFFIX)),
+    ];
+
+    if let Some(target_triple) = current_target_triple() {
+        candidates.push(
+            root.join("target")
+                .join(target_triple)
+                .join("debug")
+                .join(exporter_bin_name()),
+        );
+        candidates.push(
+            root.join("target")
+                .join(target_triple)
+                .join("debug")
+                .join(format!("cap{}", std::env::consts::EXE_SUFFIX)),
+        );
+        candidates.push(root.join("target").join("debug").join(format!(
+            "cap-exporter-{target_triple}{}",
+            std::env::consts::EXE_SUFFIX
+        )));
     }
 
     candidates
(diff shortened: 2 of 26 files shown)


richiemcilroy:
hey @greptileai please re-review the PR

richiemcilroy:
hey @greptileai please re-review the PR

PR #1913, opened 2026-06-13
fix(macos): release leaked window delegate Box and ObjC class on window close

ManthanNimodiya:
## Summary

setup() in apps/desktop/src-tauri/src/platform/macos/delegates.rs (used for traffic-light positioning on Editor, ScreenshotEditor, Settings, Upgrade and ModeSelect windows) leaked on every call:

1. A boxed WindowState (holding a live Window handle) was never freed.
2. A brand new Objective-C delegate class with a randomized name was registered every call and never disposed, along with its alloc'd instance.

Since closing these windows destroys and later rebuilds the underlying webview window, every open/close cycle permanently leaked a window handle plus a new ObjC class. This code has been in place since Nov 2024.

## Changes

- Register the delegate class once (OnceLock) and reuse it for every window; only alloc a new instance per window.
- windowWillClose: now drops the boxed window state and releases the delegate instance's alloc reference.

## Test plan
- [x] cargo dev build compiles cleanly
- [x] Dev app launches and runs without crashes

<h3>Greptile Summary</h3>

This PR fixes two memory leaks that accumulated on every open/close cycle of the traffic-light delegate windows: `Box<WindowState<R>>` was never freed because `Box::into_raw` was never balanced, and a new Objective-C class (with a randomized name) was registered on every `setup()` call and never disposed.

- The delegate class is now registered exactly once via `std::sync::OnceLock` and reused for every window, eliminating the per-call ObjC class leak.
- `windowWillClose:` now drops the boxed `WindowState`, nulls the ivar, restores the previous delegate, and calls `release` to balance the `new` allocation — all with careful ordering to avoid use-after-free on late callbacks.
- `with_window_state` now guards against the null `app_box` left behind after close, addressing the previously-flagged SIGSEGV path for enqueued post-close delegate callbacks.

The change correctly fixes two long-standing leaks and handles the cleanup ordering carefully; the one open question (class registration failure) affects only an edge-case startup path and leaves the rest of the delegate logic intact.

The core leak fixes are well-structured: the `OnceLock` class-registration correctly avoids duplicate registration across `setup()` calls, the `windowWillClose:` cleanup sequence (drop box → null ivar → restore delegate → release) is ordered to prevent dangling-reference callbacks, and the null guard in `with_window_state` closes the previously-flagged crash path. The only concern is a robustness gap in the `OnceLock` initializer: a panic there would poison the lock for the process lifetime, but this requires an ObjC class name collision that is very unlikely in practice and does not affect the correctness of the leak fix itself.

No files require special attention beyond the single changed file.

| Filename | Overview |
|----------|----------|
| apps/desktop/src-tauri/src/platform/macos/delegates.rs | Fixes two memory leaks: `Box<WindowState>` is now properly dropped in `windowWillClose:`, and the delegate ObjC class is registered once via `OnceLock` instead of per-call. The `app_box` null guard addresses the previously-flagged SIGSEGV path for late callbacks. One robustness gap remains: if `ClassDecl::new` returns `None` the `expect` panic inside `OnceLock::get_or_init` poisons the lock, permanently breaking all subsequent `setup()` calls for the process lifetime. |

</details>

1. `apps/desktop/src-tauri/src/platform/macos/delegates.rs`, line 87-96 ([link](https://github.com/capsoftware/cap/blob/c713ff1e3cd354f65b2a6b6167cb027ddfe2c511/apps/desktop/src-tauri/src/platform/macos/delegates.rs#L87-L96)) 

   <a href="#">P1</a> **`with_window_state` dereferences `app_box` without a null check**

   `on_window_will_close` sets `app_box` to `null` (line 132) and then calls `release`, potentially freeing the delegate. If any subsequent delegate message (e.g. a `windowDidResize:` notification enqueued on the main thread before the window fully closed) fires after `windowWillClose:` returns, `with_window_state` will cast a null pointer to `&mut WindowState<R>` — a SIGSEGV that `suppress_delegate_panic`'s `catch_unwind` will not catch. Adding a null guard here would prevent the crash without changing normal-path behavior.

   `````markdown
   Path: apps/desktop/src-tauri/src/platform/macos/delegates.rs
   Line: 87-96

   Comment:
   **`with_window_state` dereferences `app_box` without a null check**

   `on_window_will_close` sets `app_box` to `null` (line 132) and then calls `release`, potentially freeing the delegate. If any subsequent delegate message (e.g. a `windowDidResize:` notification enqueued on the main thread before the window fully closed) fires after `windowWillClose:` returns, `with_window_state` will cast a null pointer to `&mut WindowState<R>` — a SIGSEGV that `suppress_delegate_panic`'s `catch_unwind` will not catch. Adding a null guard here would prevent the crash without changing normal-path behavior.

   `````
   </details>
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(macos): cargo fmt and make windowWil..."](https://github.com/capsoftware/cap/commit/9f4bb63d54872105c53fdd3b351c536cd997f709) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37397018)</sub>

### DIFF
--- a/apps/desktop/src-tauri/src/platform/macos/delegates.rs
+++ b/apps/desktop/src-tauri/src/platform/macos/delegates.rs
@@ -369,8 +369,12 @@ pub fn setup<R: Runtime>(window: Window<R>, controls_inset: LogicalPosition<f64>
         fn get_or_register_delegate_class<R: Runtime>() -> &'static Class {
             static CLASS: std::sync::OnceLock<&'static Class> = std::sync::OnceLock::new();
             CLASS.get_or_init(|| {
+                if let Some(existing) = Class::get("CapWindowDelegate") {
+                    return existing;
+                }
+
                 let mut decl = ClassDecl::new("CapWindowDelegate", class!(NSObject))
-                    .expect("CapWindowDelegate class already registered");
+                    .expect("failed to register CapWindowDelegate");
 
                 decl.add_ivar::<id>("window");
                 decl.add_ivar::<*mut c_void>("app_box");
--- a/apps/desktop/src-tauri/src/platform/macos/delegates.rs
+++ b/apps/desktop/src-tauri/src/platform/macos/delegates.rs
@@ -368,7 +368,7 @@ pub fn setup<R: Runtime>(window: Window<R>, controls_inset: LogicalPosition<f64>
         // be baked into the shared class for all `R`.
         fn get_or_register_delegate_class<R: Runtime>() -> &'static Class {
             static CLASS: std::sync::OnceLock<&'static Class> = std::sync::OnceLock::new();
-            *CLASS.get_or_init(|| {
+            CLASS.get_or_init(|| {
                 let mut decl = ClassDecl::new("CapWindowDelegate", class!(NSObject))
                     .expect("CapWindowDelegate class already registered");
 
--- a/apps/desktop/src-tauri/src/platform/macos/delegates.rs
+++ b/apps/desktop/src-tauri/src/platform/macos/delegates.rs
@@ -135,9 +135,9 @@ pub fn setup<R: Runtime>(window: Window<R>, controls_inset: LogicalPosition<f64>
                 // that was leaked via `Box::into_raw` when this delegate was created.
                 let app_box: *mut c_void = *this.get_ivar("app_box");
                 if !app_box.is_null() {
-                    drop(Box::from_raw(app_box as *mut WindowState<R>));
                     let this_mut = this as *const Object as *mut Object;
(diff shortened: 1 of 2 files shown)


ManthanNimodiya:
@greptileai please re-review

ManthanNimodiya:
@tembo please re-review

ManthanNimodiya:
@greptileai please re-review

ManthanNimodiya:
@tembo please re-review

ManthanNimodiya:
hey @richiemcilroy, looks like the tips are still pending on algora's side, this one and the one from #1865 both seem to need you to click through, please have a look when you get chance
thank you 🙏🏻

PR #1917, opened 2026-06-16
desktop: gate instant uploads with a cached health probe

Peter7896:
/claim #73

This keeps the upload-speed and upload-health logic on one bounded path for Instant mode instead of doing a real object upload in the middle of the recording flow.

What changed:
- add a desktop upload-health probe with a bounded payload, short timeout, and a 10 minute cache
- use the cached result to cap Instant recording resolution before recording starts
- add a web upload-health route that measures uploadability without leaving probe objects behind in storage
- surface the health state in the desktop UI and add focused desktop/web coverage around tiering and byte limits

Follow-up hardening:
- validate the server-reported receivedBytes before calculating throughput
- subtract a same-route HEAD round trip estimate from the POST timing when it is safe to do so, reducing high-latency underestimation
- short-circuit overlapping refresh probes so concurrent startup, interval, manual, or post-recording checks do not race the cache
- avoid clearing the refresh spinner from non-refresh status reads
- make the Content-Length check an explicit fast path while keeping stream-side byte counting as the enforcement boundary

Why this shape:
- recording start only consumes cached state, so we do not wait on a fresh speed check after the user has already started recording
- the health route exercises the upload path without introducing orphaned probe files or cleanup requirements
- resolution gating stays conservative and only narrows Instant mode when the measured path says uploads are weak
- the RTT adjustment avoids punishing high-latency links when actual upload throughput is still healthy

Validation:
- git diff --check
- pnpm exec biome check on touched TypeScript files
- pnpm --dir apps/desktop exec vitest run src/utils/upload-health.test.ts
- pnpm --dir apps/web exec vitest run __tests__/unit/desktop-upload-health.test.ts
- cargo fmt --manifest-path apps/desktop/src-tauri/Cargo.toml --check
- cargo test -j 1 -p cap-desktop upload_health --lib

### DIFF
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -45,6 +45,7 @@ mod tray;
 mod update_project_names;
 mod updates;
 mod upload;
+mod upload_health;
 pub mod web_api;
 mod window_exclusion;
 mod window_position_persistence;
@@ -4844,6 +4845,8 @@ pub async fn run(recording_logging_handle: LoggingHandle, logs_dir: PathBuf) {
             recording::restart_recording,
             recording::delete_recording,
             recording::take_screenshot,
+            upload_health::get_upload_health_status,
+            upload_health::refresh_upload_health_status,
             recording::import_current_desktop_background,
             recording::list_cameras,
             recording::get_camera_formats,
@@ -5217,6 +5220,7 @@ pub async fn run(recording_logging_handle: LoggingHandle, logs_dir: PathBuf) {
             app.manage(FinalizingRecordings::default());
             app.manage(updates::UpdatesState::default());
             updates::spawn_background_loop(app.clone());
+            app.manage(upload_health::UploadHealthCache::default());
 
             #[cfg(unix)]
             {
--- a/apps/desktop/src-tauri/src/recording.rs
+++ b/apps/desktop/src-tauri/src/recording.rs
@@ -1631,7 +1631,7 @@ pub async fn start_recording(
                 notify_recording_start_failed(&app, &error);
                 return Err(error);
             };
-            let instant_mode_max_resolution = if auth.is_upgraded() {
+            let mut instant_mode_max_resolution = if auth.is_upgraded() {
                 general_settings
                     .map_or(cap_recording::PRO_INSTANT_MODE_MAX_RESOLUTION, |settings| {
                         settings.instant_mode_max_resolution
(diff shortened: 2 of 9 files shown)


Peter7896:
Rebased this branch onto current `main` in 4d6c8498a and resolved the merge conflict in `apps/desktop/src/routes/(window-chrome)/new-main/index.tsx` by keeping the upstream updater import together with the upload-health imports. The PR diff remains scoped to the desktop upload health probe path.

Validation:
- `pnpm --dir apps/desktop exec vitest run src/utils/upload-health.test.ts` -> 6 passed
- `pnpm --dir apps/web exec vitest run __tests__/unit/desktop-upload-health.test.ts` -> 3 passed
- `pnpm exec biome check apps/desktop/src/routes/(window-chrome)/new-main/index.tsx apps/desktop/src/utils/upload-health.ts apps/desktop/src/utils/upload-health.test.ts apps/web/__tests__/unit/desktop-upload-health.test.ts apps/web/app/api/desktop/[...route]/root.ts apps/web/app/api/desktop/[...route]/uploadHealth.ts`
- `git diff --check origin/main...HEAD`

`cargo check -p cap-desktop` is still blocked in this local Windows environment before checking this crate because the FFmpeg headers are not installed (`libavutil/avutil.h` missing in `ffmpeg-sys-next`).

Peter7896:
Rebased this branch onto current `main` and resolved the desktop conflicts by retaining the upstream updater state, teleprompter/window controls, and icon imports alongside the upload-health additions. The branch is mergeable again.

Validation:
- `cargo fmt --manifest-path apps/desktop/src-tauri/Cargo.toml --check`
- desktop upload-health tests -> 6 passed
- web upload-health tests -> 3 passed
- focused Biome check -> passed
- `git diff --check` -> passed

Current head: `94e86bb1d6d58a4803b35b3a500c14516f8696f5`.

MyTH-zyxeon:
Exact-head review of `94e86bb1d6d58a4803b35b3a500c14516f8696f5` against the open $500 issue #73 acceptance path. I traced the startup/interval UI calls through the Tauri command, cache, recording-start cap, and authenticated web probe.

1. **[P1] A probe already in flight is not stopped when recording starts.** `refresh_upload_health_status` checks `is_recording_active_or_pending()` once, before acquiring `cache.probe`, and then `run_probe` can spend up to 2 seconds on HEAD plus 8 seconds on POST without another state check or cancellation. A startup/5-minute/manual probe can therefore pass the guard, then continue consuming upload bandwidth after the user starts recording. That is the exact interval issue #73 says to avoid, and it can also distort the recording upload the probe is meant to protect. Re-check after acquiring the probe guard and make recording start cancel/join any in-flight probe (or have `run_probe` select on a recording-state cancellation token); add a race test where recording becomes pending after the first guard but before/during POST.

2. **[P1] The cold HEAD duration can be subtracted from a warm POST and materially overstate upload speed.** Both requests share the reqwest client, so the first startup HEAD may include DNS/TCP/TLS setup while the following POST reuses the connection. `upload_elapsed_after_rtt` subtracts that entire cold HEAD elapsed from the POST whenever at least 50 ms remains. For example, a 500 ms cold HEAD followed by a 700 ms POST becomes a 200 ms upload, inflating the 256 KiB result from roughly 3 Mbps to roughly 10 Mbps and selecting 1920 instead of 1280. Warm the connection first and measure a second same-connection HEAD, or cap/subtract only a separately measured warm request/response overhead; cover cold-connect > POST-overhead in the timing tests.

Positive control: the bounded streaming endpoint, `receivedBytes` validation, in-flight mutex, stale-cache refusal, and recording-start resolution cap close several earlier failure modes. The current head is mergeable, but the visible checks only show security checks plus a failing Vercel status; the focused Rust/desktop/web tests listed in the PR are not current-head CI evidence yet.

Run-Id: `run-20260810T020256Z-misa3-revenue-cycle-cap1917-review`
Trace-Id: `3ACC15F8-806D-49BF-B323-198DF6EFED4B`
Requester: million-dollar-revenue-cycle service-account `million-dollar-revenue-cycle`
Implementer: MISA 3 bot ID `1516725819517567077`

PR #1927, opened 2026-06-19
fix(web): gate thumbnail and view-count endpoints and tighten CORS

richiemcilroy:
Adds canView gating to the thumbnail and view-count endpoints so private videos are not exposed, and limits CORS credentials to trusted origins instead of reflecting any origin.

<h3>Greptile Summary</h3>

This PR closes two related security gaps: private video thumbnails and view-counts were readable by any caller, and the CORS layer was reflecting any `Origin` header back with `Access-Control-Allow-Credentials: true`, enabling credential-read CSRF from arbitrary origins.

- **Auth gating**: both `/api/thumbnail` and `/api/analytics` now run `canView` through the `VideosPolicy` effect before touching any video data; private, password-protected, and non-existent videos return 404 to unauthorized callers.
- **CORS hardening**: `getHeaders` now only sets `Access-Control-Allow-Origin: <origin>` + `credentials: true` for origins that pass `isTrustedCredentialedOrigin` (localhost, `cap.so`, `cap.link`, their subdomains, and the configured `WEB_URL`); all other origins receive the wildcard header without credentials.

Both security fixes are correctly implemented and the existing error-handling paths are preserved; safe to merge.

The CORS credential fix and policy gating are both logically sound. The `withPublicPolicy` pre-check runs before the DB query so access is denied before any data is fetched. The `Exit.isFailure` + empty-result guard in both routes covers all failure modes. No regressions introduced to existing public-video or authenticated-user flows.

| Filename | Overview |
|----------|----------|
| apps/web/utils/helpers.ts | Adds isTrustedCredentialedOrigin to gate credentials-bearing CORS responses to known origins; fixes the previous bug of reflecting any caller's origin with Allow-Credentials: true |
| apps/web/app/api/thumbnail/route.ts | Adds canView policy gate before serving thumbnail URLs; correct use of Exit.isFailure + empty-result check, though the policy itself issues a second DB query after canView's own repo.getById call |
| apps/web/app/api/analytics/route.ts | Adds canView policy gate before returning view-count data; 404 responses omit CORS headers (consistent with pre-existing behavior for this endpoint) |

</details>

`````markdown

---

### Issue 1 of 2
apps/web/utils/helpers.ts:30-32
`.cap.link` subdomains are not covered by the trusted-origin check, while `.cap.so` subdomains are. If any `*.cap.link` host needs to make credentialed requests (e.g., a future `share.cap.link` embed), it would silently fall through to the wildcard/no-credentials path. Either add the parallel subdomain check or document the intentional asymmetry.

```suggestion
	if (host === "localhost" || host === "cap.so" || host === "cap.link")
		return true;
	if (host.endsWith(".cap.so") || host.endsWith(".cap.link")) return true;
```

### Issue 2 of 2
apps/web/app/api/thumbnail/route.ts:35-40
**Double DB round-trip per request**

`videosPolicy.canView(id)` internally calls `repo.getById(videoId)` to evaluate visibility/ownership (see `buildCanView` in `VideosPolicy.ts`), and then the `Effect.promise` here issues a second `SELECT * FROM videos` for the same row. For the thumbnail route the second query returns the full record (needed for `decodeStorageVideo`), so it can't easily be eliminated — but it's worth being aware that every thumbnail request now hits the DB twice. The analytics route only selects `{ id }` on the second pass so the redundancy there is even more pronounced.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(web): gate thumbnail and view-count ..."](https://github.com/capsoftware/cap/commit/2a624e0f24339150fd327147161eaf973c0ae85e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38614756)</sub>

### DIFF
--- a/apps/web/app/api/analytics/route.ts
+++ b/apps/web/app/api/analytics/route.ts
@@ -1,5 +1,12 @@
+import { db } from "@cap/database";
+import { videos } from "@cap/database/schema";
+import { provideOptionalAuth, VideosPolicy } from "@cap/web-backend";
+import { Policy, Video } from "@cap/web-domain";
+import { eq } from "drizzle-orm";
+import { Effect, Exit } from "effect";
 import type { NextRequest } from "next/server";
 import { getVideoAnalytics } from "@/actions/videos/get-analytics";
+import * as EffectRuntime from "@/lib/server";
 
 const parseRangeParam = (value: string | null) => {
 	if (!value) return undefined;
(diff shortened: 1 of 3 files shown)


richiemcilroy:

PR #1978, opened 2026-07-04
Harden A/V sync CI, cursor timing, and Windows hardware encoding

richiemcilroy:
Follow-up to #1977, addressing the nightly sync-tests failure plus two field-reported defects.

## 1. Nightly sync-matrix flake (run 28697093459, Windows)

The randomized case `random/2/video-14fps-.../41ts` failed the absolute-pts check with 0.285s of error. Root cause is a harness expectation bug, not a pipeline bug: the case's drop lottery dropped the first 3 nominal frames, so the first frame the pipeline ever saw was sent at t=0.216s. The pipeline (by design) zeroes each track's timeline at its first delivered frame — cross-track alignment comes from the per-track `start_time` persisted by the recorder — so every muxed pts carried a structural 0.216s offset against the test's absolute origin, plus ~70ms of encoder-queue jitter on the shared runner. The relative-structure check (the sync-bearing one) passed.

Fix: the absolute check now measures against the origin-normalized sent timeline (first sent frame ⇔ first muxed pts), and the absolute tolerance is floored at the relative tolerance since it contains the same per-frame jitter. Verified locally with the failing seed (`CAP_SYNC_MATRIX_SEED=1783145408979603400`): all 28 cases pass; random/2's error drops from 285ms structural to pure scheduler jitter.

## 2. Cursor visibly lagging the video

The default cursor-smoothing spring (tension 470, friction 70) chases the raw position at the *same* timestamp, and a spring tracking a moving target trails it by `friction/tension` at steady state — a systematic **149ms** cursor delay behind the video, identical in preview and export. (The old 500ms click look-ahead used to mask this near clicks; its reduction to 140ms in `529a71580` exposed it.)

Fix: the spring's target now samples the raw path `friction/tension` ahead per active profile (149ms default, 40ms drag, ~75ms click-window), one-pole-smoothed across profile switches so the target never pops. Each regime cancels its own lag, so the smoothed cursor lands where the real cursor was at render time while keeping the smoothing aesthetic. The drawn cursor icon still samples un-led time. Locked by `smoothed_cursor_tracks_moving_target_without_lag`, which fails at ~149ms of trail without the compensation (verified by reverting).

Clock domains, pause excision, and preview/export parity were audited and are correct; remaining known offsets are sub-frame (≤21ms cross-track start snap vs raw cursor epochs, ~8–16ms poll quantization) and are documented for follow-up rather than touched here.

## 3. Windows studio recordings showing a solid green display

A solid green display with intact editor background is the render of zeroed YUV (Y=U=V=0 through the BT.601 shader). The editor decodes with software FFmpeg on Windows (deterministic on every machine), which localizes the reported machine-dependent green to encode time: the default Windows studio path picks a vendor hardware encoder (`h264_nvenc`/`h264_amf`/`h264_qsv`/`h264_mf`) with **no output validation and no runtime fallback** — unlike the non-fragmented MediaFoundation path, which validates and falls back to software. Some driver stacks open a session successfully and then emit zeroed frames with no error surfaced.

Fix: on Windows, every hardware H264 candidate must now pass a pre-flight round trip — encode a short neutral-gray clip, decode it with the software decoder, verify the gray survives — before it can be selected, in both the muxed and standalone builder paths. Failures log a warning and fall through the existing priority list, terminating at `libx264`, so NVIDIA/AMD/Intel/CPU-only machines all end up on a proven-working encoder. Results are cached per (encoder, resolution, mode); `CAP_DISABLE_ENCODER_SELF_TEST=1` is the escape hatch. macOS is exempt (single-vendor VideoToolbox, and a measured ~2.5s session+encode cost would tax recording start for a failure mode never observed there); Linux records via libx264 already.

The verifier rejects both zeroed (green) and black round-trip output; covered by unit tests plus a full round trip through a real encoder.

## 4. Capture gaps landing on a segment cut were collapsed (multi-second desync)

Found by running the randomized matrix at nightly scale locally (40 cases): two case shapes failed deterministically with 2.4–3.1s relative pts errors. Box-level analysis of the produced fragments showed the mechanism: the DASH/mov fragmenter anchors each fragment's `tfdt` at the accumulated duration of the previous fragment, and the **last sample of a fragment takes its packet duration verbatim** — which we always synthesized as the nominal frame period. When a capture gap (static screen) crosses a segment cut, the first post-gap frame is pulled back onto the pre-gap timeline (one frame period after the last pre-gap frame) and its content plays during the gap; everything muxed after it is placed correctly, leaving that frame seconds out of sync and the viewer seeing post-gap content during the hold. Curated gap cases never caught this because their gap frame is not a keyframe, so the muxer never cuts there.

Fix: `EncoderBase` now holds one packet and stamps it with the real dts delta to its successor before writing (only when the duration was synthesized), so the fragmenter's accumulated timeline includes the gap and the pre-gap frame correctly holds through it. Locked by `gap_crossing_segment_cut_preserves_post_gap_pts`, which fails on the old code with the exact production signature (post-gap frame at 0.396s instead of 3.0s). This covers the macOS/Windows/Linux studio fragmented paths and instant mode, which all sit on `SegmentedVideoEncoder`.

A second hardening in the same class: the anomaly tracker confirmed real >2s gaps only via arrival spacing at the mux loop, which encoder backpressure can defeat (pre-gap backlog and post-gap frame drain back-to-back). A frame can never be stamped ahead of the wall clock, so forward jumps landing at-or-behind it are now accepted as real gaps regardless of arrival bunching; future-stamped jumps still collapse as glitches. Locked by `bunched_real_gap_behind_wall_clock_is_accepted` / `future_stamped_jump_is_still_collapsed`.

With both fixes, the full 62-case matrix passes on the failing seed (`CAP_SYNC_MATRIX_SEED=1783149530744270000`), the previously failing cases dropping from 2.4–3.1s errors to ~65ms scheduler jitter.

<h3>Greptile Summary</h3>

This PR hardens recording sync, cursor timing, and Windows H.264 encoder selection. The main changes are:

- Adds Windows hardware encoder round-trip self-tests before selection.
- Caches self-test results by the encoder session inputs that shape output.
- Fails closed when decoded self-test frames use unexpected pixel formats.
- Holds one video packet so real durations survive segment-cut gaps.
- Updates timestamp gap handling for bunched post-gap frames.
- Normalizes sync-matrix absolute PTS checks to the first delivered frame.
- Adds cursor phase-lead compensation for spring smoothing.

This looks safe to merge.

- No blocking issues found in the changed code.

| Filename | Overview |
|----------|----------|
| crates/enc-ffmpeg/src/video/h264.rs | Adds Windows hardware H.264 self-tests and tightens cache keys plus decoded-format validation. |
| crates/enc-ffmpeg/src/base.rs | Adds one-packet holdback so synthesized durations can be replaced with real DTS deltas. |
| crates/enc-ffmpeg/src/mux/segmented_stream.rs | Retries init-segment notification and adds coverage for gaps crossing segment cuts. |
| crates/recording/src/output_pipeline/core.rs | Uses wall-clock elapsed time to distinguish real capture gaps from future-stamped jumps. |
| crates/recording/tests/sync_matrix.rs | Compares absolute PTS values against the origin-normalized sent timeline. |
| crates/rendering/src/cursor_interpolation.rs | Adds spring lead compensation so smoothed cursor motion tracks the raw path timing. |

<sub>Reviews (3): Last reviewed commit: ["Include pixel format in the self-test ca..."](https://github.com/capsoftware/cap/commit/37d2ef97577db62294b8492b71d54f4d36b9caa9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41786911)</sub>

**Context used:**

- Context used - CLAUDE.md ([source](https://app.greptile.com/cap/github/capsoftware/cap/-/custom-context?memory=9a906542-f1fe-42c1-89a2-9f252d96d9f0))
- Context used - AGENTS.md ([source](https://app.greptile.com/cap/github/capsoftware/cap/-/custom-context?memory=27801409-c24c-4476-9c6c-180f1ef0a7f2))

### DIFF
--- a/crates/enc-ffmpeg/src/base.rs
+++ b/crates/enc-ffmpeg/src/base.rs
@@ -13,6 +13,14 @@ pub struct EncoderBase {
     first_pts: Option<i64>,
     last_frame_pts: Option<i64>,
     last_written_dts: Option<i64>,
+    /// One-packet reorder buffer: a packet is written once its successor is
+    /// known so a synthesized duration can be replaced with the real dts
+    /// delta. Fragmenting muxers place each fragment at the accumulated
+    /// duration of the previous one, and the last sample of a fragment uses
+    /// the packet duration verbatim — a nominal duration there collapses any
+    /// capture gap that lands on a fragment cut, playing post-gap content
+    /// during the gap. The bool records whether the duration was synthesized.
+    held_packet: Option<(Packet, bool)>,
 }
 
 impl EncoderBase {
@@ -23,6 +31,7 @@ impl EncoderBase {
             last_frame_pts: None,
             stream_index,
             last_written_dts: None,
+            held_packet: None,
         }
     }
 
@@ -133,7 +142,8 @@ impl EncoderBase {
                 _ => {}
             }
 
-            if self.packet.duration() <= 0
+            let duration_synthesized = self.packet.duration() <= 0;
+            if duration_synthesized
                 && let Some(duration) = nominal_packet_duration(
                     output.stream(self.stream_index).unwrap().time_base(),
                     encoder.frame_rate(),
@@ -161,7 +171,18 @@ impl EncoderBase {
             }
 
             self.last_written_dts = self.packet.dts();
-            self.packet.write_interleaved(output)?;
+
+            let current = std::mem::replace(&mut self.packet, Packet::empty());
+            if let Some((mut previous, previous_synthesized)) = self.held_packet.take() {
+                if previous_synthesized
+                    && let (Some(prev_dts), Some(cur_dts)) = (previous.dts(), current.dts())
+                    && cur_dts > prev_dts
+                {
+                    previous.set_duration(cur_dts - prev_dts);
+                }
+                previous.write_interleaved(output)?;
+            }
+            self.held_packet = Some((current, duration_synthesized));
         }
 
         Ok(())
@@ -174,7 +195,13 @@ impl EncoderBase {
     ) -> Result<(), ffmpeg::Error> {
         encoder.send_eof()?;
 
-        self.process_packets(output, encoder)
+        self.process_packets(output, encoder)?;
+
+        if let Some((previous, _)) = self.held_packet.take() {
+            previous.write_interleaved(output)?;
+        }
+
+        Ok(())
     }
 }
 
--- a/crates/enc-ffmpeg/src/mux/segmented_stream.rs
+++ b/crates/enc-ffmpeg/src/mux/segmented_stream.rs
@@ -328,7 +328,6 @@ impl SegmentedVideoEncoder {
         frame: frame::Video,
         timestamp: Duration,
     ) -> Result<(), QueueFrameError> {
-        let is_first_frame = self.segment_start_time.is_none();
         let segment_start = match self.segment_start_time {
             Some(start) => start,
             None => {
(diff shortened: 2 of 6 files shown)


richiemcilroy:

richiemcilroy:

PR #2017, opened 2026-07-16
Add Chinese-first short form captions

haoneowu:
Adds a Chinese-first local Whisper default and short-form caption segmentation. The shared segmenter preserves word timestamps, is used by both editor transcription flows, and includes focused tests for Chinese length, pauses, English fallback, and legacy captions.\n\nValidated with: Vitest (4 tests), Biome, and desktop TypeScript type checking.

### DIFF
--- a/NEOCAP.md
+++ b/NEOCAP.md
@@ -1,0 +1,36 @@
+# NeoCap
+
+## Metadata
+
+| Field | Value |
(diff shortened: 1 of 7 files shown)


haoneowu:
Created against upstream by mistake; closing immediately.

PR #2020, opened 2026-07-18
Make Cap fully operable by agents

richiemcilroy:
Adds an additive Effect /api/v1, comprehensive Cap CLI and local stdio MCP surface, secure browser/PKCE authentication with OS-backed credentials, Agent Skill installers, server-enforced idempotency, lifecycle cleanup, and dashboard-equivalent account, library, organization, and developer operations. Includes release-binary Docker E2E coverage across MySQL, MinIO, media processing, permissions, authentication, CLI/MCP parity, security, and regression gates.

<h3>Greptile Summary</h3>

This PR adds a comprehensive agent API surface to Cap: a new `/api/v1` Effect-based HTTP handler, PKCE browser authentication, OS-backed CLI credential storage, an MCP stdio interface, server-enforced idempotency via MySQL `FOR UPDATE` transactions, and a full suite of account/library/organization/developer operations.

- All three issues flagged in the prior review have been resolved: `lastUsedAt` is now updated in the auth middleware (with a 5-minute throttle), the Rust callback listener returns a 405 response for non-GET requests, and LIKE search terms are now escaped with `escapeAgentLikePattern` and `ESCAPE '!'`.
- Two new P2 observations: the cron cleanup route's string-length guard before `timingSafeEqual` can diverge from byte length for multi-byte UTF-8 characters (causing a throw instead of a 401), and the `agentApiOperations` cleanup only deletes terminal states, leaving any workflow-crashed records in `\"running\"` state indefinitely.

Safe to merge — the three previously-flagged defects are all addressed, auth and idempotency logic are sound, and the two remaining observations are non-blocking edge cases.

The core security paths (PKCE flow, token hashing, scope enforcement, idempotency correctness) are well-implemented. The cron timingSafeEqual edge case can only produce a 500 for malformed Authorization headers and does not allow authentication bypass. The stuck-operation cleanup gap requires a double failure scenario before it manifests operationally.

apps/web/lib/agent-api-cleanup.ts and apps/web/app/api/cron/cleanup-agent-api/route.ts contain the two minor issues noted in the comments.

| Filename | Overview |
|----------|----------|
| apps/web/app/api/v1/[...route]/route.ts | New 8,604-line agent API handler with comprehensive capability system, scope-guarded mutations, and idempotency. One P2 concern: stuck async operations in running/queued state are never reaped by the cleanup job. |
| apps/web/lib/agent-api-cleanup.ts | Handles expired auth codes, idempotency records, and access tokens correctly, but agentApiOperations deletion only targets terminal states, leaving orphaned running/queued operations to accumulate indefinitely. |
| apps/web/app/api/cron/cleanup-agent-api/route.ts | Correctly uses timingSafeEqual and rejects missing CRON_SECRET, but string-length pre-check can diverge from byte-length for multi-byte UTF-8 characters, causing timingSafeEqual to throw. |
| packages/web-backend/src/AgentAuth.ts | New auth middleware with proper SHA-256 token hashing, 5-minute lastUsedAt refresh throttle, scope validation, and legacy key path. All three issues from the previous review are resolved. |
| apps/web/lib/agent-write.ts | Server-enforced idempotency using MySQL FOR UPDATE locks — insert-then-lock pattern correctly prevents concurrent duplicate mutations and provides deterministic replay. |
| apps/web/lib/agent-token.ts | PKCE authorization code exchange is single-use, tokens are hashed before storage, and the revoke endpoint correctly scopes by principal ID. |
| apps/web/workflows/agent-cap-operation.ts | Step-based durability pattern; try/catch transitions operations to failed state on any error, but if failOperation itself fails the record stays in running state permanently. |
| apps/web/app/cli/authorize/page.tsx | Re-validates all form fields server-side, rate-limits per user, and only issues codes for loopback redirect URIs. CSRF covered by Next.js server action origin checking. |
| packages/database/schema.ts | Adds agent API tables with appropriate indexes. Tokens stored as SHA-256 hashes, not plaintext. |
| apps/cli/src/agent_auth.rs | Previously-flagged non-GET callback path now correctly sends a 405 Method Not Allowed response. PKCE flow and state validation look correct. |

<sub>Reviews (5): Last reviewed commit: ["fix: validate developer app logo URLs"](https://github.com/capsoftware/cap/commit/13b8c565b2d294e12504b5ec2c0f4285f24b3aaa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45382622)</sub>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 2.9.4",
+ "bitflags 2.13.1",
  "bytes",
  "futures-core",
  "futures-sink",
@@ -52,7 +52,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "base64 0.22.1",
- "bitflags 2.9.4",
+ "bitflags 2.13.1",
  "bytes",
  "bytestring",
  "derive_more 2.0.1",
@@ -192,6 +192,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
 name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -251,7 +262,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed7572b7ba83a31e20d1b48970ee402d2e3e0537dcfe0a3ff4d6eb7508617d43"
 dependencies = [
  "alsa-sys",
- "bitflags 2.9.4",
+ "bitflags 2.13.1",
  "cfg-if",
  "libc",
 ]
@@ -273,7 +284,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0f2a1bb052857d5dd49572219344a7332b31b76405648eabac5bc68978251bcd"
 dependencies = [
  "android-properties",
- "bitflags 2.9.4",
+ "bitflags 2.13.1",
  "cc",
  "jni 0.22.4",
  "libc",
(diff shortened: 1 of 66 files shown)


richiemcilroy:

richiemcilroy:

richiemcilroy:

ringotokens-commits:
I found one security-sensitive gap in the new agent confirmation model.

The API advertises many write/destructive capabilities as requiring `confirmation: "user"` (for example `signOutAllDevices` is destructive), but the server-side enforcement in `requireUserConfirmedRequest` only checks whether the request includes `X-Cap-Confirmation: user`. The first-party CLI only adds that header after prompting, which is good for the happy path, but the bearer token holder can also send the same header directly from any HTTP client. That means a custom client, compromised MCP process, or injected agent with a scoped token can self-confirm protected operations, so the server is not actually enforcing a distinct user confirmation step.

If confirmation is intended as a security boundary, I think this needs a server-verifiable proof of confirmation instead of a spoofable header: e.g. a short-lived confirmation challenge minted by the dashboard/native client and bound to `{tokenId, operation, idempotencyKey/body hash}`, or else the capability metadata/errors should be renamed to make it explicit that this is only client-side UX friction. Otherwise users may grant broad scopes believing destructive actions still require a separate user-mediated approval.

richiemcilroy:

PR #2025, opened 2026-07-23
fix: exclude iOS-only mobile dev task from dev:windows

Ectropy:
dev:windows doesn't work on windows since it currently doesn't exclude the recently added iOS simulator step. For windows development only we should filter this package out.

@cap/mobile#dev boots an iOS simulator and cannot run on Windows; its POSIX env-var prefix also fails to parse under cmd, so the task exits immediately and turbo tears down every sibling dev task, killing dev:windows within seconds of starting. Filter the package out the same way dev:web already does.

<h3>Greptile Summary</h3>

Updates the Windows development command to avoid launching the iOS-only mobile task.
- Adds the existing `@cap/mobile` Turbo exclusion filter to `dev:windows`.

The PR appears safe to merge.

The focused script change excludes the Windows-incompatible mobile development task without altering other development-task selection or runtime behavior.

| Filename | Overview |
|----------|----------|
| package.json | The Windows development script now excludes `@cap/mobile`, matching the established negative-filter pattern and preventing the incompatible iOS simulator task from terminating the run. |

<sub>Reviews (1): Last reviewed commit: ["fix: exclude iOS-only mobile dev task fr..."](https://github.com/capsoftware/cap/commit/80214b8380a26dbfe8c6cd721ee25f48d8f4fb88) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46569461)</sub>


richiemcilroy:
thank you!

PR #2027, opened 2026-07-23
feat(chrome-extension): open the recorder in Chrome's side panel on non-injectable pages

ManthanNimodiya:
Follow-up to the "recorder opens in a separate window" confusion from the support thread.

On pages the extension can't inject into (new tab, chrome://, Web Store), the Cap icon now docks the recorder in Chrome's side panel instead of a floating popup window. It survives tab switches, the icon toggles it, and normal websites keep the in-page panel exactly as shipped. The popup window stays as a logged last resort.

Implementation notes: sidePanel.open() must ride the click gesture (any timer voids it), so the decision happens synchronously in the action handler and the fallback delivery path avoids timed retries. The standalone panel reports open/closed to the service worker so teardown reaches it and only one recorder shows at a time.

Also includes two bug fixes found while testing, both reproducible on main (can split into a separate PR):
- After every extension update/reload, orphaned content scripts blocked re-injection via a stale global flag, leaving open tabs unable to show Cap UI until refreshed. Injection now takes over the tab, mounts a fresh overlay module, and sweeps stale UI.
- A recording that dies without a terminal status left the offscreen doc pinned in a phantom "recording" phase: stop was a no-op and every tab's bar froze until browser restart. Stop now clears zombie phases and the idle-close interval doubles as a watchdog.

Tested manually on Chrome/macOS (side panel open/toggle/persistence, normal sites unchanged incl. the cap.so dashboard flow, recording from the side panel, orphan recovery after reload). Typecheck, unit tests, and builds pass. sidePanel needs Chrome 116 = existing minimum_chrome_version. Firefox has no sidePanel API; the Firefox port handles this separately.

<h3>Greptile Summary</h3>

The PR adds a side-panel recorder for non-injectable Chrome pages and improves extension recovery behavior.
- Adds the side-panel permission, manifest entry, lifecycle messages, toggle handling, and popup fallback.
- Reworks content-script reinjection to replace orphaned extension generations and stale overlay UI.
- Clears zombie offscreen recording phases during stop handling and idle watchdog checks.

The PR is not yet safe to merge because the first post-restart click can lose the user gesture required to open the side panel and fall back to the confusing popup window.

The startup-race fix waits for a callback-based context query inside the click listener, allowing the gesture-gated side-panel open to execute too late on the exact worker-restart path it is intended to repair.

**Files Needing Attention:** apps/chrome-extension/src/background/service-worker.ts

| Filename | Overview |
|----------|----------|
| apps/chrome-extension/src/background/service-worker.ts | Adds side-panel lifecycle and toggle orchestration, but the startup-state fix awaits an asynchronous context query before the gesture-gated open call. |
| apps/chrome-extension/src/popup/main.tsx | Reports standalone panel lifecycle to the worker and handles close broadcasts. |
| apps/chrome-extension/src/content/bootstrap.ts | Replaces stale bootstrap generations and removes orphaned overlay UI during reinjection. |
| apps/chrome-extension/src/offscreen/recorder.ts | Detects and clears recording phases that no longer have an active capture operation. |
| apps/chrome-extension/public/manifest.json | Enables the Chrome side-panel permission and configures the standalone recorder page. |
| apps/chrome-extension/src/shared/types.ts | Adds service-worker request variants for standalone panel lifecycle events. |

`````markdown
### Issue 1
apps/chrome-extension/src/background/service-worker.ts:1931
**Startup wait loses click gesture**

When the first icon click after an MV3 worker restart arrives before `standaloneFlagReady` settles, this `await` suspends the action handler until the asynchronous `getContexts` callback returns. `sidePanel.open` then runs outside the required click gesture, rejects, and the fallback path ultimately opens the legacy popup instead of the side panel.

---

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["fix(extension): await standaloneFlagRead..."](https://github.com/capsoftware/cap/commit/b99e560fbb2f40a04a23cbe9fe4883d23514b91f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46733291)</sub>

### DIFF
--- a/apps/chrome-extension/public/manifest.json
+++ b/apps/chrome-extension/public/manifest.json
@@ -31,26 +31,55 @@
 		"identity",
 		"offscreen",
 		"scripting",
+		"sidePanel",
 		"storage",
 		"tabCapture"
 	],
-	"host_permissions": ["http://*/*", "https://*/*", "file:///*"],
+	"host_permissions": [
+		"http://*/*",
+		"https://*/*",
+		"file:///*"
+	],
 	"content_scripts": [
 		{
-			"matches": ["http://*/*", "https://*/*", "file:///*"],
-			"js": ["assets/content-bootstrap.js"],
+			"matches": [
+				"http://*/*",
+				"https://*/*",
+				"file:///*"
+			],
+			"js": [
+				"assets/content-bootstrap.js"
+			],
 			"run_at": "document_idle"
 		}
 	],
 	"web_accessible_resources": [
 		{
-			"resources": ["icons/*", "content/overlay.js", "welcome.html"],
-			"matches": ["http://*/*", "https://*/*", "file:///*"]
+			"resources": [
+				"icons/*",
+				"content/overlay.js",
+				"welcome.html"
+			],
+			"matches": [
+				"http://*/*",
+				"https://*/*",
+				"file:///*"
+			]
 		},
 		{
-			"resources": ["camera-preview.html", "popup.html"],
-			"matches": ["http://*/*", "https://*/*", "file:///*"],
+			"resources": [
+				"camera-preview.html",
+				"popup.html"
+			],
+			"matches": [
+				"http://*/*",
+				"https://*/*",
+				"file:///*"
+			],
 			"use_dynamic_url": true
 		}
-	]
+	],
+	"side_panel": {
+		"default_path": "popup-window.html"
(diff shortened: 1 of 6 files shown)


ManthanNimodiya:
@greptileai

ManthanNimodiya:
@greptileai

PR #2028, opened 2026-07-24
fix: preserve loom csv retry quota

### LINKED ISSUE #2010: CSV Loom import: rate-limit tokens consumed before dedupe check, making retries import nothing
## Summary

Bulk Loom import via CSV cannot complete for large libraries. The per-user rate
limit empties partway through the CSV, all remaining rows fail, and — because
the rate-limit check runs before the already-imported check — retrying the same
CSV spends the entire rate-limit budget on rows that already succeeded and
imports zero new videos.

## Environment

- Cap Pro (web), CSV import at /dashboard (Loom importer)
- Library: ~1,500 Loom videos, importing in CSVs under the 500-row cap
- Google Drive storage integration enabled

## Steps to reproduce

1. Import a CSV with more rows than the `rl_loom_import_per_user` window allows
   (~60 in my case).
2. First ~60 rows import; every subsequent row fails with
   `"Too many Loom imports started. Please wait a few minutes, then try again."`
3. Wait, then re-submit the same CSV to pick up the failed rows.
4. Result: `{"success":false,"importedCount":0, ...}` — every row fails with the
   same rate-limit error. No new videos import.

## Cause (from source)

In `apps/web/actions/loom.ts`, the CSV loop calls `isRateLimited()` per row
*before* `importLoomVideoForOwner()`, and the duplicate check
(`"This Loom video has already been imported."`) lives inside
`importLoomVideoForOwner()`. So on a retry, every already-imported row consumes
a rate-limit token at the gate and never reaches the dedupe short-circuit. For
a 500-row CSV where 60 succeeded, the retry burns the whole window on rows
1-60 plus rows that will fail anyway, and the user can never make progress
without manually trimming completed rows out of the CSV.

The loop also has no pacing or backoff, so a single large CSV always drains
the bucket and fails the tail in one shot.

## Suggested fixes (any one of these unblocks large imports)

1. **Reorder the checks:** run the dedupe lookup before `isRateLimited()` so
   already-imported rows return early without spending rate-limit budget.
   Retrying the same CSV then becomes safe and idempotent.
2. **Queue instead of fail:** when the limit trips mid-CSV, pause and resume
   rather than failing the remaining rows (the job is already async
   server-side).
3. **Surface the budget:** return remaining quota / retry-after in the response
   so the UI can tell users how many rows to submit and when to resume.

## Actual response sample

```json
{"success":false,"importedCount":0,"failedCount":10,"results":[
  {"rowNumber":112,"userEmail":"…","success":false,
   "error":"Too many Loom imports started. Please wait a few minutes, then try again."},
  …
],"error":"No Loom videos were imported."}
```

Happy to test a fix — I have ~1,440 videos still to import and an active Loom
account to reproduce against.

tylergibbs1:
## what changed

csv retries now run the existing loom duplicate lookup before charging a rate limit token. rows with an existing cap return the same duplicate result, while new and stale rows still go through the limiter.

the limiter moved into `importLoomVideoForOwner` for the csv path. this keeps the lookup and stale row handling in one place and avoids another database query.

the regression test covers a duplicate row followed by a new row and checks that only the new import spends quota.

## why

a retry of a partially imported csv spent its full budget on rows that had already succeeded. large imports could never reach the first unfinished row unless someone edited completed rows out of the csv.

closes #2010

## testing

- `pnpm --filter @cap/web test`
- `pnpm --dir apps/web exec next typegen`
- `pnpm exec tsc -b apps/web/tsconfig.json --pretty false`
- `pnpm exec biome check apps/web/actions/loom.ts apps/web/__tests__/unit/loom-import.test.ts`

<h3>Greptile Summary</h3>

This PR preserves Loom CSV retry quota by moving per-row rate-limit checks behind the existing duplicate lookup.
- Already-imported Loom rows now return the existing duplicate result without consuming quota.
- New and stale rows continue through the rate limiter before import processing.
- A regression test verifies that a duplicate row followed by a new row charges quota only for the new import.

The PR appears safe to merge, with duplicate retries bypassing quota consumption while importable rows remain rate-limited.

The changed ordering preserves existing duplicate and stale-record behavior, keeps the single-video path unchanged, and applies the CSV limiter before any new import begins.

| Filename | Overview |
|----------|----------|
| apps/web/actions/loom.ts | Moves the CSV rate-limit callback into the shared import helper after live-duplicate detection while retaining checks for new and stale imports. |
| apps/web/__tests__/unit/loom-import.test.ts | Adds coverage confirming that duplicate CSV retries preserve quota and a subsequent new row still starts its import. |

<sub>Reviews (1): Last reviewed commit: ["fix: preserve loom csv retry quota"](https://github.com/capsoftware/cap/commit/46fe6434c204995be3df42c39751d22a00ac9dd0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46904886)</sub>

### DIFF
--- a/apps/web/__tests__/unit/loom-import.test.ts
+++ b/apps/web/__tests__/unit/loom-import.test.ts
@@ -640,6 +640,99 @@ describe("importFromLoom", () => {
 		expect(valuesMock).not.toHaveBeenCalled();
 	});
 
+	it("skips rate limit checks for csv rows that were already imported", async () => {
+		whereMock.mockImplementation((conditions: unknown) => {
+			const serializedConditions = JSON.stringify(conditions);
+
+			if (serializedConditions.includes("sourceId")) {
+				return Promise.resolve(
+					serializedConditions.includes("loom-existing123")
+						? [{ videoId: "existing-video" }]
+						: [],
+				);
+			}
+
+			return withLimit([{ userId: "member-123", email: "member@example.com" }]);
+		});
+
+		const fetchMock = vi.mocked(fetch);
+		fetchMock.mockImplementation(async (input) => {
+			const url = typeof input === "string" ? input : input.toString();
+
+			if (url.includes("/transcoded-url")) {
+				return {
+					ok: true,
+					status: 200,
+					text: async () =>
+						JSON.stringify({ url: "https://cdn.loom.com/video.mp4" }),
+				} as Response;
+			}
+
+			if (url === "https://www.loom.com/graphql") {
+				return {
+					ok: true,
+					json: async () => ({
+						data: { getVideo: { name: "Imported video" } },
+					}),
+				} as Response;
+			}
+
+			if (url.includes("/v1/oembed")) {
+				return {
+					ok: true,
+					json: async () => ({ duration: 42, width: 1920, height: 1080 }),
+				} as Response;
+			}
+
+			throw new Error(`Unexpected fetch: ${url}`);
+		});
+
+		const { importFromLoomCsv } = await import("@/actions/loom");
+
+		const result = await importFromLoomCsv({
+			orgId: "org-1" as never,
+			rows: [
+				{
+					rowNumber: 2,
+					loomUrl: "https://www.loom.com/share/loom-existing123",
+					userEmail: "member@example.com",
+				},
+				{
+					rowNumber: 3,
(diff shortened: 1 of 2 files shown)


richiemcilroy:
thanks for your contribution!

Issue #1981, opened 2026-07-06
Unauthenticated Cross-User AI Metadata Disclosure via Missing Access Check on /api/video/ai


geo-chen:
reported on 4 June 2026 https://github.com/CapSoftware/Cap/security/advisories/GHSA-5m2p-v3wg-3gm6 - no response:

### Summary

The GET /api/video/ai endpoint retrieves AI-generated metadata (title, summary, chapters) for any video by ID without verifying whether the requesting user has permission to access that video. Any authenticated user can supply an arbitrary video ID and read the AI-generated content of another user's private recordings.

### Details

The handler at apps/web/app/api/video/ai/route.ts (GET, lines 13-169) fetches the target video directly from the database using only the caller's authenticated session and the videoId query parameter:

```
const result = await db()
  .select()
  .from(videos)
  .where(eq(videos.id, videoId));
```

No ownership check and no call to the VideosPolicy canView policy are made before returning the video's metadata fields. The response body includes:

- metadata.aiTitle
- metadata.summary
- metadata.chapters
- metadata.aiGenerationStatus

The correct access-guarded path used by every other video-reading endpoint is getByIdForViewing, which runs the full canView policy (owner check, org/space membership, public flag, password, email restriction). This endpoint bypasses all of that.

### PoC

Preconditions: User A has a private video (public=false) with AI metadata populated. User B has no membership in User A's org or space and cannot view the video at /s/{videoId}.

1. Authenticate as User B and obtain a session token (cookie next-auth.session-token).

2. Note the video ID of User A's private recording (vid001vidtest00 in the test below). In practice an attacker would enumerate IDs or obtain one via a notification link or a shared URL they intercept.

3. Issue:

```
curl -s "http://<host>/api/video/ai?videoId=vid001vidtest00" \
  -H "Cookie: next-auth.session-token=<User B token>"
```

4. Response (200 OK):

```json
{
  "processing": false,
  "title": "Secret Report Title",
  "summary": "This is a confidential summary",
  "chapters": [],
  "aiGenerationStatus": "COMPLETE"
}
```

User B receives the AI-generated title and summary of a video they cannot open. The /s/vid001vidtest00 share page correctly shows "This video is private" to User B, confirming there is no legitimate access path.

The same endpoint will also attempt to enqueue AI generation on behalf of the requester when aiGenerationStatus is SKIPPED and transcriptionStatus is COMPLETE, consuming the video owner's AI credits without their knowledge.

### Impact

[...]

PR #1926, opened 2026-06-19
fix(web): require canView on transcript and AI metadata endpoints

richiemcilroy:
Gates the transcript, translate-transcript, available-translations and AI-metadata endpoints behind the same canView policy used elsewhere, so private and password-protected videos' content is no longer readable by unauthorized callers. Also rate-limits the translation endpoint.

<h3>Greptile Summary</h3>

This PR closes an authorization gap by wrapping the transcript, translate-transcript, available-translations, and AI-metadata endpoints in the same `canView` policy already used elsewhere, and adds Vercel Firewall rate limiting to the translation endpoint.

- All four endpoints now use `Policy.withPublicPolicy(videosPolicy.canView(videoId))` via `provideOptionalAuth`; policy failures and not-found cases both return the same opaque message to avoid leaking video existence for private/password-protected content.
- `apps/web/lib/rate-limit.ts` introduces a shared `isRateLimited` helper backed by `@vercel/firewall`; it is best-effort (fails open on error) and no-ops outside production, matching the pattern already established in `actions/collections/password.ts`.
- The `RATE_LIMIT_IDS` constants beyond `TRANSLATE_TRANSCRIPT` are placeholders for future rules; only the translation endpoint is wired in this PR.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change only adds access-control gates and a rate-limit check; existing authorized callers are unaffected.

The authorization fix is applied consistently across all four endpoints using the same well-established pattern. The rate-limit helper is best-effort (fails open), so there is no availability risk. No logic regressions were found in the happy paths.

No files require special attention.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| apps/web/actions/videos/get-transcript.ts | Adds canView policy gate via withPublicPolicy before the video DB fetch; access-denied and not-found cases both surface as the same "Video not found" response (information-safe). |
| apps/web/actions/videos/get-available-translations.ts | Applies the same canView-gated Effect pattern; storage listing remains inside the try/catch so policy failures are cleanly separated from storage errors. |
| apps/web/actions/videos/translate-transcript.ts | Adds both rate limiting (Vercel Firewall, IP-based) and canView policy before the expensive Groq translation path; ordering is correct (fast-reject first). |
| apps/web/app/api/video/ai/route.ts | Adds canView policy on top of the existing auth guard; previously any authenticated user could read AI metadata for any video by ID — now restricted to users with view permission. |
[...]

Issue #1993, opened 2026-07-09
bug(self-hosting): Chrome extension login fails on self-hosted instances (missing CAP_CHROME_EXTENSION_ID)


jerryno6:
### Bug Description
When self-hosting Cap using the default Docker Compose templates (`docker-compose.yml` or `docker-compose.coolify.yml`), logging in via the Chrome extension fails with a `400 Bad Request` or an `HttpApiDecodeError` indicating `redirectUri` validation issues.

### Steps to Reproduce
1. Deploy Cap using the self-hosted Docker templates.
2. Configure the Chrome Extension's Cap URL to point to your self-hosted instance (e.g., `https://cap.yourdomain.com`).
3. Click "Log In" in the extension.
4. The extension attempts to authenticate against `/api/extension/auth/start` but receives a `400 Bad Request` / failure.
```
  Could not reach Cap extension auth at https://cap.yourdomain.com. Start the local web server with pnpm dev:extension and make sure the extension Options Cap URL is
  https://cap.hapytek.com. Redirect URI: https://fefjaffcodfiogbbngmjkcjpbpclbdcp.chromiumapp.org/. 400
```

### Technical Context
In `packages/env/server.ts`, the environment variable `CAP_CHROME_EXTENSION_ID` is used to validate the authorized Chrome extension ID. However:
- `CAP_CHROME_EXTENSION_ID` is not mapped to the container in `docker-compose.yml` or `docker-compose.coolify.yml` under the `cap-web` service.
- The parameter is undocumented and missing from the env template files (e.g., `docker-compose.coolify.env.example`), making it impossible for self-hosted instances to authorize Chrome extension logins out-of-the-box.

PR #2029, opened 2026-07-25
fix(web): enforce video access policy before creating comments (#1982)

### LINKED ISSUE #1982: Authenticated Users Can Post Comments on Private Videos Without Access Permission
reported on 4 June 2026 https://github.com/CapSoftware/Cap/security/advisories/GHSA-5r3r-v2f5-wqwh - no response:

### Summary

The POST /api/video/comment endpoint accepts a videoId in the request body and inserts a comment attributed to the authenticated caller without checking whether the caller has permission to view or interact with the target video. Any authenticated user can inject comments into any other user's private recordings.

### Details

The handler at apps/web/app/api/video/comment/route.ts (POST, lines 11-87) authenticates the session to obtain a userId, then immediately inserts the comment into the database:

```typescript
const userId = user?.id || User.UserId.make("anonymous");

// ... no video access check ...

const newComment = {
  id: Comment.CommentId.make(nanoId()),
  authorId: userId,
  type: type,
  content: content,
  videoId: Video.VideoId.make(videoId),
  timestamp: timestamp || null,
  parentCommentId: parentCommentIdSanitized || null,
  createdAt: new Date(),
  updatedAt: new Date(),
};

await db().insert(comments).values(newComment);
```

There is no lookup of the video to check public status, ownership, or org/space membership before the insert. The delete path (apps/web/app/api/video/comment/delete/route.ts) correctly checks that the comment belongs to the requesting user (and(eq(comments.id, ...), eq(comments.authorId, user.id))), but the creation path has no equivalent guard.

### PoC

Preconditions: User A has a private video (public=false). User B has no membership in User A's organization and cannot view the video at /s/{videoId}.

1. Authenticate as User B and obtain a session token.

2. POST to the comment endpoint with User A's private video ID:

```
curl -s -X POST "http://<host>/api/video/comment" \
  -H "Cookie: next-auth.session-token=<User B token>" \
  -H "Content-Type: application/json" \
  -d '{"type":"text","content":"injected comment","videoId":"vid001vidtest00"}'
```

3. Response (200 OK):

```json
{
  "id": "z5ef9735wsv0h7g",
  "authorId": "tcrhw9md2124491",
  "type": "text",
  "content": "injected comment",
  "videoId": "vid001vidtest00",
  "timestamp": null,
  "parentCommentId": null,
  "createdAt": "2026-06-03T22:34:57.420Z",
  "updatedAt": "2026-06-03T22:34:57.420Z",
  "authorName": "Anonymous"
}
```

4. The comment is now stored in the database with User B's authorId and will appear if the video is later made public or if the owner views their notifications.

The comment-notification pipeline (apps/web/app/api/email/new-comment/route.ts) is also triggered, meaning the video owner receives a notification email for each injected comment.

### Impact

Any authenticated user can write comments into another user's private video without any access to that video. Beyond the authorization bypass itself, this allows an attacker to spam the video owner with comment notification emails (low-cost notification abuse), inject content into a private video's comment thread that will appear if the video is later shared, and probe whether a guessed video ID belongs to a real video (200 vs no such video response).

DPS0340:
Fixes #1982 (GHSA-5r3r-v2f5-wqwh).

## Problem

The `newComment` server action authenticates the caller, then inserts the comment without ever checking whether that user is allowed to access the target video:

```ts
const user = await getCurrentUser();
if (!user) throw new Error("User not authenticated");
// ...no video access check...
await db().insert(comments).values(newComment);
```

`videoId` is a client-supplied argument to a server action, so being signed in is the only barrier. Any authenticated user can post comments — or emoji reactions, or threaded replies — into **any** video by ID, including another user's private recording. The comment then shows up in the owner's activity feed and triggers a notification to them via `createNotification`.

Worth noting the same file already treats deletion correctly: `deleteComment` verifies `authorId` before removing anything. Creation was the gap.

## Fix

Gate the insert behind the existing `VideosPolicy.canView` policy before any write happens.

I deliberately did not invent a new access-check shape — this mirrors the pattern already used in this exact directory by `getTranscript` (`apps/web/actions/videos/get-transcript.ts`), and it is the same policy the `/api/video/ai` fix in #1926 used to close the sibling advisory #1981:

```ts
const accessExit = await Effect.gen(function* () {
  const videosPolicy = yield* VideosPolicy;
  return yield* Effect.promise(() =>
    db().select({ id: videos.id }).from(videos).where(eq(videos.id, videoId)),
  ).pipe(Policy.withPublicPolicy(videosPolicy.canView(videoId)));
}).pipe(provideOptionalAuth, EffectRuntime.runPromiseExit);

if (Exit.isFailure(accessExit) || accessExit.value.length === 0) {
  throw new Error("Video not found");
}
```

Because it reuses `canView`, it inherits the full existing access model for free — owner, org membership, space membership, public videos, password-protected videos, and email-domain restrictions — rather than approximating it with an ownership check that would have broken legitimate commenting on shared videos.

The error message is deliberately `"Video not found"` rather than `"Access denied"`, so this does not become an oracle for probing which video IDs exist.

## Verification

- `biome check apps/web/actions/videos/new-comment.ts` — clean.
- `tsc --noEmit` on `apps/web`: the remaining diagnostics on this file (TS6305, plus TS2488/TS18046 on the Effect generator) are **not introduced by this change** — the untouched `get-transcript.ts` emits the identical set, since it uses the same pattern and the workspace `tsconfig` project references aren't built in a plain `tsc` run. Verified by diffing the two files' diagnostics.
- `@cap/web-domain`, `@cap/web-backend` and `@cap/database` build clean via turbo.

**Diff: 1 file, +22/-2.**

### DIFF
--- a/apps/web/actions/videos/new-comment.ts
+++ b/apps/web/actions/videos/new-comment.ts
@@ -4,7 +4,7 @@ import { db } from "@cap/database";
 import { getCurrentUser } from "@cap/database/auth/session";
 import { nanoId } from "@cap/database/helpers";
 import { comments, videos } from "@cap/database/schema";
-import { provideOptionalAuth, VideosPolicy } from "@cap/web-backend";
+import { makeCurrentUserLayer, VideosPolicy } from "@cap/web-backend";
 import type { ImageUpload } from "@cap/web-domain";
 import { Comment, Policy, type Video } from "@cap/web-domain";
 import { eq } from "drizzle-orm";
@@ -55,7 +55,13 @@ export async function newComment(data: {
 				.where(eq(videos.id, videoId))
 				.limit(1),
 		).pipe(Policy.withPublicPolicy(videosPolicy.canView(videoId)));
-	}).pipe(provideOptionalAuth, EffectRuntime.runPromiseExit);
+	}).pipe(
+		// This action already required auth above, so provide the user we have
+		// rather than `provideOptionalAuth`, which would re-run getServerSession()
+		// and re-query `users`.
+		Effect.provide(makeCurrentUserLayer(user)),
+		EffectRuntime.runPromiseExit,
+	);
 
 	if (Exit.isFailure(accessExit)) {
 		throw new Error("Video not found");
--- a/apps/web/actions/videos/new-comment.ts
+++ b/apps/web/actions/videos/new-comment.ts
@@ -3,11 +3,15 @@
 import { db } from "@cap/database";
 import { getCurrentUser } from "@cap/database/auth/session";
 import { nanoId } from "@cap/database/helpers";
-import { comments } from "@cap/database/schema";
+import { comments, videos } from "@cap/database/schema";
+import { provideOptionalAuth, VideosPolicy } from "@cap/web-backend";
 import type { ImageUpload } from "@cap/web-domain";
-import { Comment, type Video } from "@cap/web-domain";
+import { Comment, Policy, type Video } from "@cap/web-domain";
+import { eq } from "drizzle-orm";
+import { Effect, Exit } from "effect";
 import { revalidatePath } from "next/cache";
 import { createNotification } from "@/lib/Notification";
+import * as EffectRuntime from "@/lib/server";
 
 export async function newComment(data: {
 	content: string;
@@ -37,6 +41,22 @@ export async function newComment(data: {
 	if (!content || !videoId) {
 		throw new Error("Content and videoId are required");
 	}
+
+	// The caller controls `videoId`, so being signed in is not enough — without
+	// this check any authenticated user can insert comments into a video they
+	// cannot view, including someone else's private recording.
+	const accessExit = await Effect.gen(function* () {
+		const videosPolicy = yield* VideosPolicy;
+
+		return yield* Effect.promise(() =>
+			db().select({ id: videos.id }).from(videos).where(eq(videos.id, videoId)),
+		).pipe(Policy.withPublicPolicy(videosPolicy.canView(videoId)));
+	}).pipe(provideOptionalAuth, EffectRuntime.runPromiseExit);
+
+	if (Exit.isFailure(accessExit) || accessExit.value.length === 0) {
+		throw new Error("Video not found");
+	}
+
 	const id = Comment.CommentId.make(nanoId());
 
 	const newComment = {
--- a/apps/web/actions/videos/new-comment.ts
+++ b/apps/web/actions/videos/new-comment.ts
@@ -49,11 +49,19 @@ export async function newComment(data: {
 		const videosPolicy = yield* VideosPolicy;
 
 		return yield* Effect.promise(() =>
-			db().select({ id: videos.id }).from(videos).where(eq(videos.id, videoId)),
+			db()
+				.select({ id: videos.id })
+				.from(videos)
+				.where(eq(videos.id, videoId))
+				.limit(1),
 		).pipe(Policy.withPublicPolicy(videosPolicy.canView(videoId)));
 	}).pipe(provideOptionalAuth, EffectRuntime.runPromiseExit);
 
-	if (Exit.isFailure(accessExit) || accessExit.value.length === 0) {
+	if (Exit.isFailure(accessExit)) {
+		throw new Error("Video not found");
+	}
+
+	if (accessExit.value.length === 0) {
 		throw new Error("Video not found");
 	}
 


DPS0340:
@tembo please review

DPS0340:
Adopted — pushed. Split the branches so the type guard narrows cleanly, matching `get-transcript.ts`. Also added `.limit(1)` to the lookup while I was in there, since `videos.id` is unique.

DPS0340:
@tembo please re-review

DPS0340:
Adopted — pushed.

Verified the redundancy before changing it: `provideOptionalAuth` is literally

```ts
const user = yield* getCurrentUser;   // getServerSession() + select from users
return yield* user.pipe(Option.match({
  onNone: () => app,
  onSome: (user) => app.pipe(Effect.provide(makeCurrentUserLayer(user))),
}));
```

so on a path that has already established `user` it re-runs both the session lookup and the `users` query to arrive at the same layer. Since this action throws on `!user` before the policy check, the `onNone` branch is unreachable here — `provideOptionalAuth` was pure overhead.

Now `Effect.provide(makeCurrentUserLayer(user))`, which matches what `dashboard/folder/[id]/page.tsx` and `dashboard/spaces/[spaceId]/page.tsx` already do.

Small bonus: this also removed a `TS18046 'accessExit.value' is of type 'unknown'` diagnostic — providing the layer concretely gives the compiler a narrower env type than the `provideOptionalAuth` wrapper did.

DPS0340:
@tembo please re-review

DPS0340:
@tembo please re-review

DPS0340:
@richiemcilroy — nine PRs from me are sitting green and I don't want them to read as a wall, so here they are ordered by review cost. Every one is `MERGEABLE`, rebased on current `main` (`5b81ee5b`), and every `@tembo` note is answered.

**Cheapest three — one file each, no behavioural surface beyond the fix:**

| PR | size | what |
|---|---|---|
| #2036 | +6/-0, 2 files | self-hosting: AI provider keys never reached the web container |
| #2035 | +29/-0, 4 files | self-hosting: same for `CAP_CHROME_EXTENSION_ID` (#1993) |
| #2030 | +30/-8, 1 file | `getVideoAnalytics` server action wasn't gated on the video access policy |

**The access-policy set — these are one theme, and reviewing #2029 first makes the rest fast:**

- **#2029** (+36/-2) — comments could be created on videos the caller can't view (#1982)
- **#2031** (+45/-3) — space/folder video listing didn't verify space membership
- **#2032** (+185/-2) — `/api/video/transcribe/status` ungated
- **#2038** (+149/-18) — download permission granted the owner's whole org while *view* required an explicit share; this aligns them

**Two standalone:**

- **#2034** (+78/-3) — ProRes swscale output didn't match the declared full colour range
- **#2040** (+145/-0) — 3 of 13 `RATE_LIMIT_IDS` were declared but never called, so those endpoints had no rate limit; also adds a guard so a new unused id fails a test instead of silently shipping

Two notes so review time goes to the code and not to context:

- **The `Vercel — Authorization required to deploy` red mark is not mine.** It's the fork-preview authorisation, and it's present on merged PRs too — #1966 carries the identical failure. Everything else (Socket, Superagent, Contributor trust) is green.
- **#2038 had a real defect that @tembo caught** — a signature change with a stale call site in `apps/web/app/s/[videoId]/page.tsx` that would have broken the build. Fixed in `b988753`, and I re-grepped the other call sites rather than assuming that was the only one.

No rush on any individual one — happy for you to take the three cheap ones and leave the rest, if that's the useful split.

DPS0340:
@richiemcilroy — **#2021 already fixes this, and it got there first. Please review that one instead of this.**

I opened this on 25 July without noticing @eponomariova-dotcom's PR from 20 July. They are the same fix in the same file, and theirs is the more finished one:

| | #2021 (theirs, 20 Jul) | #2029 (mine, 25 Jul) |
|---|---|---|
| file | `apps/web/actions/videos/new-comment.ts` | same |
| size | +33/-2 | +36/-2 |
| review rounds | 8 `@tembo` notes, all addressed | 3 |

Their version also independently caught the subtlety I did — `canView` returns `true` for a nonexistent `videoId` by design, so gating a no-op still lets a bogus id reach the insert; both fetch the row under the policy. On one point theirs is **better**: `@tembo` pointed out that a bare `Exit.isFailure` also swallows operational errors (DB connection) as "Video not found", and they split and logged it. Mine still collapses both.

The only real difference left is `provideOptionalAuth` (theirs) vs `Effect.provide(makeCurrentUserLayer(user))` (mine). I chose the latter to avoid re-running `getServerSession()` and re-querying `users` when the action has already authenticated — a small saving, and **not worth preferring a later, less-reviewed PR over an earlier one**. If it's useful it's a two-line follow-up to theirs, and I'm happy to send that once #2021 lands.

I'll close this in favour of #2021 unless you'd rather I keep it open as a fallback — say the word and it's gone.

Sorry for the duplicated review surface. I checked #1982 for existing *linked* PRs but not for open PRs referencing it, which is the check that would have caught this before I wrote anything. My other eight PRs don't overlap #2021 — I verified each touches a different file.

DPS0340:
Closing as promised — #2021 by @eponomariova-dotcom covers this and got there first.

No action needed from anyone; I'm removing it from the review queue rather than leaving it sitting as something you'd have to diff and dismiss. Nothing here is lost: if any piece turns out to be worth having after #2021 by @eponomariova-dotcom lands, I'll send it as a small follow-up against that.

PR #2038, opened 2026-07-26
fix(web): require an explicit share to download, matching the view policy

### LINKED ISSUE #2037: Download permission grants the owner's whole org, while view permission requires an explicit share
### Summary

`canUserDownloadVideo()` in `apps/web/lib/video-download-permissions.ts` treats the video's **own** organization as if it were a share target, so every member of the owner's org can download any video that org owns, whether or not it was ever shared with them.

```ts
const sharedOrgs = await db()
  .select({ organizationId: sharedVideos.organizationId })
  .from(sharedVideos)
  .where(eq(sharedVideos.videoId, videoId));

const orgIds = [orgId, ...sharedOrgs.map((org) => org.organizationId)];
//              ^^^^^ the owner's org, unconditionally
```

`orgId` comes straight from `video.orgId` at the call site in `apps/web/actions/videos/download.ts`:

```ts
const allowed = await canUserDownloadVideo({
  userId: user.id,
  ownerId: video.ownerId,
  videoId,
  orgId: video.orgId,
});
```

`video.orgId` is the organization the video *belongs to*, not an organization it was shared with. Mixing it into the same list as the `sharedVideos` rows means membership of the owner's org is sufficient, and the `sharedVideos` lookup below it becomes redundant for that org.

### Why this looks unintended rather than by design

The view policy right next door requires an **explicit** share. `buildCanView` in `packages/web-backend/src/Videos/VideosPolicy.ts` uses `orgsRepo.membershipForVideo`, which inner-joins through `sharedVideos`:

```ts
db.select({ membershipId: Db.organizationMembers.id })
  .from(Db.organizationMembers)
  .leftJoin(Db.sharedVideos, Dz.eq(Db.organizationMembers.organizationId, Db.sharedVideos.organizationId))
  .where(Dz.and(
    Dz.eq(Db.organizationMembers.userId, userId),
    Dz.eq(Db.sharedVideos.videoId, videoId),      // <- the share row must exist
  ))
```

and logs `"Explicit org/space membership found. Access granted."` when it matches. `video.orgId` is used in that policy only to look up the allowed *email domain*, never as a grant.

So the two paths disagree: a colleague in the owner's org who cannot **view** an unshared private video can still **download** it, because the download check never consults `sharedVideos` for that org — it short-circuits on `orgIds` containing `video.orgId`.

The space branch in the same function is written the stricter way, which is what makes the org branch look like an oversight: it derives space ids purely from `spaceVideos` rows for this video, and returns `false` when there are none.

### Expected

Download permission is granted by the same rule as view permission: ownership, an explicit `sharedVideos` row for an org the user belongs to, or an explicit `spaceVideos` row for a space the user belongs to.

### Reproduction (by inspection of the query, not run against a live DB)

1. User A and user B are both members of org O.
2. A records a private video; it is stored with `orgId = O` and is never shared to any org or space.
3. B calls `getVideoDownloadInfo(videoId)`.
4. `canUserDownloadVideo` builds `orgIds = [O]`, finds B's `organizationMembers` row for O, and returns `true`.
5. The same B loading the video page is refused by `buildCanView`, which finds no `sharedVideos` row.

I want to be straight about the limits of my evidence here: I traced this by reading the two queries and the call site, and I do not have a running Cap instance with a seeded database to demonstrate the divergence end to end. If `video.orgId` is intended to be a grant and the view policy is the stricter one on purpose, then this is a documentation gap rather than a bug and I would rather be told so than have it merged. `video-download-permissions.ts` currently has no test file, so nothing pins either interpretation.

I have a fix ready and will open a PR referencing this issue.

DPS0340:
## What

`canUserDownloadVideo` no longer treats the video's own organization as a share target. Download permission now requires the same thing view permission requires: ownership, an explicit `sharedVideos` row for an org the user belongs to, or an explicit `spaceVideos` row for a space they belong to.

Closes #2037.

## The problem

```ts
const orgIds = [orgId, ...sharedOrgs.map((org) => org.organizationId)];
//              ^^^^^ video.orgId, unconditionally
```

`orgId` is passed from `video.orgId` at the call site in `actions/videos/download.ts`. That is the org the video *belongs to*, not one it was shared with — so membership of the owner's org was enough to download anything that org owns, and the `sharedVideos` lookup above it was redundant for that org.

The view policy next door is stricter. `buildCanView` in `packages/web-backend/src/Videos/VideosPolicy.ts` goes through `orgsRepo.membershipForVideo`, which joins through `sharedVideos` and requires the share row to exist:

```ts
.where(Dz.and(
  Dz.eq(Db.organizationMembers.userId, userId),
  Dz.eq(Db.sharedVideos.videoId, videoId),   // the share must exist
))
```

There, `video.orgId` is used only to look up the allowed email *domain*, never as a grant.

So the two paths disagreed: a colleague in the owner's org who cannot **view** an unshared private video could still **download** it. The space branch of the same function was already written the stricter way — it derives space ids purely from `spaceVideos` rows and returns `false` when there are none — which is part of why the org branch reads as an oversight.

## The change

The org membership lookup now runs only when `sharedVideos` returned rows, and only against those org ids. As a side effect it skips a query entirely when the video has no org shares, which is the common case for a personal recording.

`orgId` becomes unused, so it is dropped from the signature and from the single call site.

## Verification

I want to be precise about what I did and did not run.

**What I ran.** I could not install the full monorepo toolchain, so I ran the committed test file, unmodified, against the real `video-download-permissions.ts` in an isolated vitest project with the same `@/lib` alias this repo's `vitest.config.ts` defines. Six tests, all passing:

- the owner is allowed without any query at all
- **a colleague in the owner's org is refused when the video was never shared** — the regression, and it also asserts only 2 queries run, not 4
- a member of an org the video *was* shared with is allowed
- shared to an org the user is not in is refused
- a member of a space the video was shared into is allowed
- a space the user is not in is refused

**Revert check.** I restored the original implementation (with `orgId` and the unconditional `orgIds`) and re-ran the same tests: exactly 2 fail. The other 4 pass under both implementations, so they confirm the change did not simply make the function refuse more things.

**Direct proof of the old behaviour.** With the original code, an unshared private video plus a colleague who is a member of the owner's org returns:

```text
CURRENT code, unshared private video, colleague in owner org -> download allowed? true
```

**What I did not do.** I have not run this against a live Cap instance with a seeded database, so the query-shape reasoning is from reading the drizzle chains rather than from observed SQL. `video-download-permissions.ts` had no test file before this, so nothing pinned either interpretation.

If `video.orgId` was meant to be a grant and the view policy is deliberately the stricter of the two, then this is a docs gap rather than a bug, and I would much rather be told that than have it merged. The test file is written so it documents whichever rule you land on.

### DIFF
--- a/apps/web/__tests__/unit/video-download-permissions.test.ts
+++ b/apps/web/__tests__/unit/video-download-permissions.test.ts
@@ -1,0 +1,128 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * `canUserDownloadVideo` issues its queries in a fixed order, so each test
+ * declares the rows every query returns, in order:
+ *
+ *   1. sharedVideos    - orgs this video was explicitly shared with
+ *   2. organizationMembers - the user's membership in one of those orgs (only when 1 is non-empty)
+ *   3. spaceVideos     - spaces this video sits in
+ *   4. spaceMembers    - the user's membership in one of those spaces (only when 3 is non-empty)
+ */
+const mocks = vi.hoisted(() => {
+	const queue: unknown[][] = [];
+	let selectCalls = 0;
+
+	const db = () => ({
+		select: () => {
+			selectCalls += 1;
+			const rows = queue.shift() ?? [];
+			// The production code awaits some chains directly and calls .limit()
+			// on others, so the tail satisfies both shapes.
+			const tail = {
+				limit: async () => rows,
+				then: (resolve: (value: unknown[]) => unknown) => resolve(rows),
+			};
+			return { from: () => ({ where: () => tail }) };
+		},
+	});
+
+	return {
+		db,
+		queue,
+		reset() {
+			queue.length = 0;
+			selectCalls = 0;
+		},
+		get selectCalls() {
+			return selectCalls;
+		},
+	};
+});
+
+vi.mock("@cap/database", () => ({ db: mocks.db }));
+
+vi.mock("@cap/database/schema", () => ({
+	organizationMembers: {
+		id: "organizationMembers.id",
+		userId: "organizationMembers.userId",
+		organizationId: "organizationMembers.organizationId",
+	},
+	sharedVideos: {
+		videoId: "sharedVideos.videoId",
+		organizationId: "sharedVideos.organizationId",
+	},
+	spaceMembers: {
+		id: "spaceMembers.id",
+		userId: "spaceMembers.userId",
+		spaceId: "spaceMembers.spaceId",
+	},
+	spaceVideos: {
+		videoId: "spaceVideos.videoId",
+		spaceId: "spaceVideos.spaceId",
+	},
+}));
+
+import { canUserDownloadVideo } from "@/lib/video-download-permissions";
+
+const request = (overrides: Record<string, string> = {}) =>
+	({
+		userId: "user-b",
+		ownerId: "user-a",
+		videoId: "video-1",
+		...overrides,
+	}) as Parameters<typeof canUserDownloadVideo>[0];
+
+beforeEach(() => {
+	mocks.reset();
+});
+
+describe("canUserDownloadVideo", () => {
+	it("allows the owner without querying anything", async () => {
+		await expect(
+			canUserDownloadVideo(request({ userId: "user-a" })),
+		).resolves.toBe(true);
+
+		expect(mocks.selectCalls).toBe(0);
+	});
+
+	it("refuses a colleague when the video was never shared", async () => {
+		// The video belongs to the owner's org, but no sharedVideos or
+		// spaceVideos row exists. Membership of the owner's org is not a grant:
+		// buildCanView requires an explicit share, and download must match it.
+		mocks.queue.push([], []);
+
+		await expect(canUserDownloadVideo(request())).resolves.toBe(false);
+
+		// With no share rows there is nothing to match a membership against, so
+		// neither membership lookup should run: 2 queries, not 4.
+		expect(mocks.selectCalls).toBe(2);
+	});
+
+	it("allows a member of an org the video was explicitly shared with", async () => {
+		mocks.queue.push(
+			[{ organizationId: "org-shared" }],
+			[{ id: "org-membership-1" }],
+		);
+
+		await expect(canUserDownloadVideo(request())).resolves.toBe(true);
+	});
+
+	it("refuses when shared to an org the user does not belong to", async () => {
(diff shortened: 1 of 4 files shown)


DPS0340:
@tembo please review

DPS0340:
Note on the red check so it isn't read as a build break: `Vercel` failed with

> Authorization required to deploy.

That is the preview-deploy authorization gate for an outside contributor, not a compile or test failure. My other open PRs on this repo show the identical status. `Contributor trust` and both Socket Security checks pass.

The verification I ran locally on this exact diff is in the PR body, including the revert check confirming the 2 new behavioural tests fail against the current implementation and the other 4 pass under both.

DPS0340:
@tembo please review

DPS0340:
Good catch, and it would have broken the build — fixed in `b988753`.

I checked the call sites with `grep -rn canUserDownloadVideo apps/web` when I made the change and only acted on `actions/videos/download.ts`. `app/s/[videoId]/page.tsx` was in that output and I missed it. Now verified exhaustively against the tracked file list:

```
$ git ls-files -z '*.ts' '*.tsx' | xargs -0 grep -ln canUserDownloadVideo
apps/web/actions/videos/download.ts
apps/web/app/s/[videoId]/page.tsx
apps/web/lib/video-download-permissions.ts
```

Both call sites now pass only `{ userId, ownerId, videoId }`.

One thing I confirmed before removing the line: `video.orgId` is still used elsewhere in that page (the `organizations` join at line 792), so dropping it from the call does not leave an unused binding.

The behavioural point matters more here than in the server action, actually — `page.tsx` is what decides whether the download button renders for a viewer, so before this change a colleague in the owner's org saw a download button on a video they could not view. Now the button follows the same rule as access.

@tembo please re-review.

DPS0340:
Closing this one first, and it is the clearest cut of the four.

**#2038 closes #2037 — but I filed #2037 myself.** Every other PR I am keeping closes an issue a real user reported (@geo-chen, @rlnt, @pulgueta, @shreyastaware). This one is me reporting a thing and then fixing my own report, which is exactly the kind of item that should not be taking up space in your review queue ahead of theirs.

The underlying observation still stands as far as I can tell: `download.ts` grants the owner's entire organisation, while *viewing* the same video requires an explicit share, so download is the looser of the two. But that is a **policy call, not a bug** — you may well intend org members to be able to download. It needs your decision before it needs my patch, and it is 149 lines across 4 files to review for something that might be working as designed.

I have left #2037 open as the question. If you say "yes, download should match view", I will reopen this against current `main`.

Trimming my open PRs from nine to five; the other three closures are for the same reason (queue size, no defect found).

PR #2036, opened 2026-07-25
fix(self-hosting): pass AI provider keys through to the web container

DPS0340:
Same wiring-gap class as #2035 / #1993, found by sweeping every var in the env schema against both compose files.

## Problem

`self-hosting.mdx` tells self-hosters to set these:

```bash
DEEPGRAM_API_KEY=your-key    # Transcription
GROQ_API_KEY=your-key        # AI summaries (preferred)
OPENAI_API_KEY=your-key      # AI summaries (fallback)
```

and `docker-compose.coolify.env.example` lists all three (lines 64, 67, 68). But **neither compose file passes any of them into `cap-web`**, so the operator sets them and nothing happens.

The failure is silent, which is what makes it worth fixing rather than just documenting. Both features check the key and return early:

```ts
// apps/web/lib/transcribe.ts:32
if (!serverEnv().DEEPGRAM_API_KEY) {
  return { success: false, message: "Missing necessary environment variables" };
}

// apps/web/lib/generate-ai.ts:32
if (!serverEnv().GROQ_API_KEY && !serverEnv().OPENAI_API_KEY) {
```

So on a self-hosted instance transcription and AI summaries just never produce output, with no configuration error surfaced to the user — they followed the docs and the feature is simply absent.

## Fix

Map the three vars in both compose files, matching how the other optional integrations (`RESEND_API_KEY`, `GOOGLE_CLIENT_ID`, `APPLE_CLIENT_SECRET`) are already wired.

Defaulting to empty via `:-` means behaviour is unchanged for anyone not using these: the vars stay unset, the schema treats them as `optional()`, and the existing early-returns keep the features off.

## How I found it

Extracted every uppercase identifier from `packages/env/server.ts` and diffed against both compose files. Most of the unmapped ones are legitimately cloud-only (`VERCEL_*`, `STRIPE_*`, `WORKOS_*`, `CLOUDFRONT_*`, `TINYBIRD_*`) and shouldn't be in a self-host template. These three are different: **the self-hosting docs explicitly instruct users to set them**, which makes the omission a bug rather than a scoping decision.

`CAP_CHROME_EXTENSION_ID` was the fourth in that category and is handled in #2035.

## Verification

Both files parse as valid YAML and all three keys resolve:

```
docker-compose.yml
   DEEPGRAM_API_KEY: '${DEEPGRAM_API_KEY:-}'
   GROQ_API_KEY: '${GROQ_API_KEY:-}'
[...]

PR #2040, opened 2026-07-26
fix(web): wire three unused rate-limit ids and guard against new ones

### LINKED ISSUE #2039: 7 of 13 RATE_LIMIT_IDS are declared but never called, so those endpoints have no rate limit
### Summary

`RATE_LIMIT_IDS` in `apps/web/lib/rate-limit.ts` declares 13 rule ids, each with a doc comment naming the abuse it guards and a suggested window. **7 of them are never referenced anywhere in the codebase**, so the endpoints they describe run with no rate limit at all.

Counting references outside the declaration file itself:

| id | referenced | endpoint it names |
| --- | --- | --- |
| `AGENT_TOKEN_EXCHANGE` | 2 | wired |
| `AGENT_AUTHORIZATION` | 2 | wired |
| `AGENT_UNLOCK` | 1 | wired |
| `AGENT_LOOM_IMPORT` | 2 | wired |
| `TRANSLATE_TRANSCRIPT` | 1 | wired |
| `DOCS_ASK` | 1 | wired |
| **`AUTH_OTP_VERIFY`** | **0** | Email OTP verification (brute-force guard) |
| **`AUTH_OTP_SEND`** | **0** | Email OTP / magic-link send (mailbomb + token-reseed) |
| **`LOOM_DOWNLOAD`** | **0** | Unauthed Loom download/convert (ffmpeg + memory DoS) |
| **`MESSENGER_MESSAGE`** | **0** | Anonymous support chat (Groq + Supermemory cost) |
| **`ANALYTICS_TRACK`** | **0** | Unauthed view tracking (Tinybird ingest + notifications) |
| **`GUEST_CHECKOUT`** | **0** | Unauthed guest checkout (Stripe object/cost abuse) |
| **`DESKTOP_LOGS`** | **0** | Unauthed desktop log to Discord forwarding (spam) |

```
$ grep -rn AUTH_OTP_VERIFY apps packages --include='*.ts' --include='*.tsx' | grep -v lib/rate-limit.ts
(no output)
```

Same for the other six.

### Why this is easy to miss

The file's own comment explains that a rule id with no matching Vercel Firewall dashboard rule **fails open**:

> An ID that has no matching dashboard rule fails OPEN (`checkRateLimit` returns `{ rateLimited: false, error: "not-found" }`) ... but it also provides no protection until the rule exists.

So there is already one silent-failure mode by design. A declared-but-uncalled id is a second one, and it is invisible from the dashboard side too: someone can create `rl_guest_checkout` in the Firewall, see it configured, and reasonably assume the endpoint is protected while no code ever calls it.

### Two of the unprotected endpoints, checked

**`apps/web/app/api/settings/billing/guest-checkout/route.ts`** — no `isRateLimited`, no `getCurrentUser`, no auth of any kind. It reads `priceId` and `quantity` from the JSON body and calls `stripe().checkout.sessions.create(...)` directly. Anyone can create unbounded Stripe checkout sessions.

**`apps/web/app/api/analytics/track/route.ts`** — no `isRateLimited`. It uses `provideOptionalAuth`, so anonymous requests are expected, and it writes to Tinybird and can trigger `createAnonymousViewNotification` / `sendFirstViewEmail`. That is the exact combination the comment describes ("Tinybird ingest + notifications").

I checked those two by reading the route files. I did not audit all seven endpoints in the same depth, and I would rather say so than imply a completeness I have not verified.

### What "wired" looks like, for reference

`apps/web/app/api/docs/ask/route.ts` is the model:

```ts
if (
  isLocallyRateLimited(ip) ||
  (await isRateLimited(RATE_LIMIT_IDS.DOCS_ASK, { headers: request.headers }))
) {
  return Response.json({ error: "Too many questions right now. Try again in a minute." }, { status: 429 });
}
```

### Expected

Either each declared id is called at the endpoint it names, or the unused ones are removed so the constant does not imply protection that is not there.

### Suggested direction

I would rather not blanket-wire seven endpoints in one PR without knowing the intent — some may be deliberately staged ahead of the dashboard rules being created. What I would suggest, and am happy to implement:

1. A cheap regression guard: a unit test that asserts every key in `RATE_LIMIT_IDS` is referenced at least once outside the declaration file. That turns "declared but never called" into a CI failure rather than something found by reading.
2. Then wire the endpoints, either all in one follow-up or one at a time, whichever you prefer to review.

I have (1) ready and will open a PR referencing this issue, since it is the part that holds regardless of what you decide about each endpoint. Tell me which of the seven you want wired and I will follow up.

DPS0340:
## What

Wires three declared-but-never-called rate-limit ids to the unauthenticated endpoints they name, and adds a test so a new unwired id fails CI instead of being found by reading.

Closes #2039.

## Why

`RATE_LIMIT_IDS` declares 13 ids, each with a doc comment naming the abuse it guards. Counting references outside the declaration file, **7 were never called**, so those endpoints ran with no limit:

```
$ git ls-files -z '*.ts' '*.tsx' | xargs -0 grep -l RATE_LIMIT_IDS.GUEST_CHECKOUT | grep -v lib/rate-limit.ts
(no output)
```

This is quiet because there is already one silent-failure mode by design — the file's own comment notes that an id with no matching Vercel Firewall dashboard rule **fails open**. A declared-but-uncalled id is a second one, and it is invisible from the dashboard side too: someone can create `rl_guest_checkout` in the Firewall, see it configured, and reasonably assume the endpoint is protected while no code ever calls it.

## What this PR wires

Three endpoints I read end to end and confirmed are unauthenticated with no existing limit:

- **`api/settings/billing/guest-checkout`** — no auth of any kind. Reads `priceId`/`quantity` from the body and calls `stripe().checkout.sessions.create(...)`. Limited before anything is read from the body.
- **`api/analytics/track`** — anonymous by design (`provideOptionalAuth`). Each accepted call is a Tinybird ingest and can fan out into `createAnonymousViewNotification` / `sendFirstViewEmail`. Exactly the cost the comment describes.
- **`api/tools/loom-download`** — unauthenticated, downloads a remote video and can run it through ffmpeg. Limited before any fetching starts.

All three follow the existing `api/docs/ask` pattern: `isRateLimited(id, { headers: request.headers })`, return 429. They inherit the helper's existing best-effort semantics, so a firewall outage or a self-hosted deploy without the Firewall behaves exactly as it does today.

## What this PR deliberately does NOT wire

`AUTH_OTP_VERIFY`, `AUTH_OTP_SEND`, `MESSENGER_MESSAGE` and `DESKTOP_LOGS`. I could not confidently locate the single right call site for each, and guessing at an auth or messaging path is not something I want to do blind in a security change. They are listed in a `KNOWN_UNWIRED` map in the test with a reason each, so:

- the suite passes today,
- a **new** unwired id fails immediately rather than joining an already-failing count,
- and removing an entry is the explicit act of wiring one.

Tell me which of the four you want next and I will follow up with the call sites.

## Testing

New `__tests__/unit/rate-limit-ids.test.ts`, 4 tests:

- every declared id is referenced somewhere outside `lib/rate-limit.ts`, except the documented `KNOWN_UNWIRED` entries
- **no stale allowance**: an id in `KNOWN_UNWIRED` that *is* now referenced fails, so the list cannot linger and mask a future regression
- **no phantom allowance**: an entry naming an id that no longer exists fails
- rule ids are unique

The scan uses `git ls-files` for its file list, so it sees exactly the tracked sources and never walks `node_modules` or `.next`.

**Verification.** I ran the test's logic against the real repo rather than a copy, since the result depends on the actual file set:

```text
files scanned: 960
unused (should be []): []
stale allowances (should be []): []
unknown allowances (should be []): []
```

**Revert check.** Reverting just the `analytics/track` wiring and re-running:

```text
after reverting ANALYTICS_TRACK wiring, unused = ["ANALYTICS_TRACK"]
```

so the test genuinely detects an unwired id rather than passing by construction.

**What I did not do.** I have not exercised these three routes against a live deploy with the Firewall rules present — `isRateLimited` returns `false` outside production, so the added branch is inert in a local run. The three ids still need matching Rate Limit rules in the Vercel Firewall dashboard for the protection to take effect, per the helper's own documentation; this PR only makes the code path exist.

### DIFF
--- a/apps/web/__tests__/unit/rate-limit-ids.test.ts
+++ b/apps/web/__tests__/unit/rate-limit-ids.test.ts
@@ -25,9 +25,14 @@ const KNOWN_UNWIRED: Record<string, string> = {
  * Every key referenced somewhere other than its own declaration.
  *
  * Uses git's own file list so the search sees exactly the tracked sources and
- * never walks node_modules or .next.
+ * never walks node_modules or .next. Memoized: the scan reads every tracked
+ * TS/TSX file, and each test in this suite needs the same answer.
  */
+let referencedCache: Set<string> | undefined;
+
 function referencedKeys(): Set<string> {
+	if (referencedCache) return referencedCache;
+
 	const tracked = execFileSync(
 		"git",
 		["ls-files", "-z", "*.ts", "*.tsx"],
@@ -43,8 +48,13 @@ function referencedKeys(): Set<string> {
 		let source: string;
 		try {
 			source = readFileSync(join(WEB_ROOT, file), "utf8");
-		} catch {
-			continue; // deleted from the worktree but still in the index
+		} catch (error) {
+			// A tracked file missing from the worktree is expected (deleted but
+			// still in the index). Anything else - a permission problem, an I/O
+			// error - would silently shrink the scan and turn this guard into a
+			// no-op, so it has to surface.
+			if ((error as NodeJS.ErrnoException).code === "ENOENT") continue;
+			throw error;
 		}
 
 		for (const key of keys) {
@@ -52,6 +62,7 @@ function referencedKeys(): Set<string> {
 		}
 	}
 
+	referencedCache = found;
 	return found;
 }
 
--- a/apps/web/__tests__/unit/rate-limit-ids.test.ts
+++ b/apps/web/__tests__/unit/rate-limit-ids.test.ts
@@ -7,6 +7,7 @@ import { RATE_LIMIT_IDS } from "@/lib/rate-limit";
 
 const WEB_ROOT = join(__dirname, "..", "..");
 const DECLARATION_FILE = "lib/rate-limit.ts";
+const TEST_DIR = "__tests__/";
 
 /**
  * Ids that are intentionally declared ahead of being wired. Each entry needs a
@@ -39,7 +40,16 @@ function referencedKeys(): Set<string> {
 		{ cwd: WEB_ROOT, encoding: "utf8", maxBuffer: 32 * 1024 * 1024 },
 	)
 		.split("\0")
-		.filter((file) => file && file !== DECLARATION_FILE);
+		.filter(
+			(file) =>
+				file &&
+				file !== DECLARATION_FILE &&
+				// Only runtime call sites count. A test that merely names an id
+				// would otherwise make an unwired endpoint look protected — and
+				// this file itself references every key, which would make the
+				// guard vacuous.
+				!file.startsWith(TEST_DIR),
+		);
 
 	const found = new Set<string>();
 	const keys = Object.keys(RATE_LIMIT_IDS);
--- a/apps/web/__tests__/unit/rate-limit-ids.test.ts
+++ b/apps/web/__tests__/unit/rate-limit-ids.test.ts
@@ -1,0 +1,96 @@
+import { execFileSync } from "node:child_process";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, it } from "vitest";
+
+import { RATE_LIMIT_IDS } from "@/lib/rate-limit";
+
+const WEB_ROOT = join(__dirname, "..", "..");
+const DECLARATION_FILE = "lib/rate-limit.ts";
+
+/**
+ * Ids that are intentionally declared ahead of being wired. Each entry needs a
+ * reason; removing one from this list and wiring the endpoint is the goal.
+ * Keeping it explicit means a NEW unwired id fails the test instead of quietly
+ * joining an already-failing count.
+ */
+const KNOWN_UNWIRED: Record<string, string> = {
+	AUTH_OTP_VERIFY: "no OTP verification route located yet (see #2039)",
+	AUTH_OTP_SEND: "no OTP send route located yet (see #2039)",
+	MESSENGER_MESSAGE: "anonymous support chat route not located yet (see #2039)",
+	DESKTOP_LOGS: "desktop log forwarding route not located yet (see #2039)",
+};
+
+/**
+ * Every key referenced somewhere other than its own declaration.
+ *
+ * Uses git's own file list so the search sees exactly the tracked sources and
+ * never walks node_modules or .next.
+ */
+function referencedKeys(): Set<string> {
+	const tracked = execFileSync(
+		"git",
+		["ls-files", "-z", "*.ts", "*.tsx"],
+		{ cwd: WEB_ROOT, encoding: "utf8", maxBuffer: 32 * 1024 * 1024 },
+	)
+		.split("\0")
+		.filter((file) => file && file !== DECLARATION_FILE);
+
+	const found = new Set<string>();
+	const keys = Object.keys(RATE_LIMIT_IDS);
+
+	for (const file of tracked) {
+		let source: string;
+		try {
+			source = readFileSync(join(WEB_ROOT, file), "utf8");
+		} catch {
+			continue; // deleted from the worktree but still in the index
+		}
+
+		for (const key of keys) {
+			if (source.includes(`RATE_LIMIT_IDS.${key}`)) found.add(key);
+		}
+	}
+
+	return found;
+}
+
+describe("RATE_LIMIT_IDS", () => {
+	it("has a unique rule id per key", () => {
+		const ids = Object.values(RATE_LIMIT_IDS);
+
+		expect(new Set(ids).size).toBe(ids.length);
+	});
+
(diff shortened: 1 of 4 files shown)


DPS0340:
@tembo please review

DPS0340:
Both taken, pushed in `7237d2a`.

**Swallowing every read error.** You're right, and it's the worse of the two because of what this test is for: a silently shrinking scan makes the guard pass while checking less, which is the exact failure mode the test exists to prevent. Now only `ENOENT` continues (tracked-but-deleted, the case the comment described); anything else rethrows. Verified both branches:

```
missing file:
  ENOENT -> continue (as designed)
directory (EISDIR):
  EISDIR -> rethrown
  caught at caller: EISDIR
```

**Memoizing.** Done. Measured against the real repo:

```
first scan 41ms, memoized call 0ms
```

so it's three fewer full walks of the 960 tracked files across the suite.

Re-ran the assertions against the actual repo after both changes — the answer depends on the real tracked file set, so I check it there rather than in a fixture:

```
unused: []
stale : []
unknown: []
```

@tembo please re-review.

DPS0340:
Right again, and this one was the most consequential of the three — pushed in `3487ea1`.

It's worse than "a test could mask an unwired id": **this file itself names every key**, so once it lands, the scan would see all 13 referenced and the guard would be permanently vacuous. It only passed review because the assertions happen to be written as `RATE_LIMIT_IDS.${key}` template interpolation rather than literal `RATE_LIMIT_IDS.FOO` text, which is a very thin thing to be relying on.

I proved the hole rather than assuming it. Staged a probe test whose only content is a reference to `DESKTOP_LOGS` (an id that is *not* wired), then ran the scan both ways:

```text
with a test-only reference to DESKTOP_LOGS (an UNWIRED id):
  scanning tests too   -> unused = []           (DESKTOP_LOGS now looks wired)
  excluding tests      -> unused = []
  stale-allowance check, scanning tests ->  ["DESKTOP_LOGS"]   (false alarm)
  stale-allowance check, excluding tests -> []
```

So the test-inclusive scan both hides a genuinely unwired id *and* fires a false "stale allowance" failure telling you to delete a `KNOWN_UNWIRED` entry that is still accurate. Probe removed afterwards.

Current state of the tracked-file set, tests excluded:

```text
files scanned (tests excluded): 856   (was 960)
unused : []
stale  : []
unknown: []
```

For the record, no id is test-only in `main` today — `AGENT_TOKEN_EXCHANGE` and `AGENT_AUTHORIZATION` appear in both tests and runtime, so nothing changes about which ids currently pass. This is purely closing the future hole.

@tembo please re-review.

DPS0340:
@richiemcilroy — **correcting myself: this duplicates your #1929, it does not complement it.**

I said on #2031 that #2040 was "not a duplicate". I was wrong, and I'd rather correct it than let it stand. I compared the actual patches afterwards:

```
#1929 (yours, 19 Jun)   analytics/track      -> isRateLimited(RATE_LIMIT_IDS.ANALYTICS_*)
                        guest-checkout       -> isRateLimited(RATE_LIMIT_IDS.GUEST_CHECKOUT)
                        loom-download        -> isRateLimited(RATE_LIMIT_IDS.*)
                        + loom.ts, messenger.ts, desktop root.ts, lib/rate-limit.ts

#2040 (mine, 26 Jul)    analytics/track, guest-checkout, loom-download  — the same three,
                        with the same helper and the same ids
```

Yours wires those three **and** three more call sites I never touched, plus the `isRateLimited` helper itself. So this PR is a strict subset of work you already had open five weeks earlier.

**The one piece that isn't in #1929** is `apps/web/__tests__/unit/rate-limit-ids.test.ts` — a test asserting every declared `RATE_LIMIT_IDS` entry has a call site, so a future unwired id fails CI instead of being found by reading the file. That was the part I actually cared about; the wiring was just what the test needed to make green.

Suggested resolution: **merge #1929, and I'll reopen the test on its own** against your wiring, once I can see which ids you consider intentionally-unused. That's a ~40-line test-only PR with no overlap. Closing this otherwise.

Apologies for three overlapping PRs in one batch (#2029, #2031, this one). The cause was a single skipped step — I checked whether the linked *issue* had a PR, but never listed open PRs touching the same *files*. I've since run that across all nine; the remaining six (#2030, #2032, #2034, #2035, #2036, #2038) are clean, and I'll run it before opening anything here again.

DPS0340:
Closing as promised — #1929 (yours) covers this and got there first.

No action needed from anyone; I'm removing it from the review queue rather than leaving it sitting as something you'd have to diff and dismiss. Nothing here is lost: if any piece turns out to be worth having after #1929 (yours) lands, I'll send it as a small follow-up against that.

Issue #1791, opened 2026-05-09
Cap transcription model fails to download


asdrubalivan:
### Description

When I try to download **any** transcription model in the captions tab of the cap editor, it fails

### Reproduction

I go to the cap editor, press "Download recommended model" and the download fails.

When I try to do it again, it looks like the result is cached so I can't trigger the download again, it just instantly fails 

### Additional Context

- Cap version:Version 0.4.84 (0.4.84)
- Operating system, version: Mac OS Tahoe 26.4.1
- Device (optional):

The debug file is attached 

[logs.txt](https://github.com/user-attachments/files/27556422/logs.txt)

DPS0340:
@asdrubalivan's attached logs pin this down, and I think **it is already fixed on `main`** — worth confirming before anyone spends time on a patch.

### What the logs show

Every failure is the same, and the timing is the giveaway:

```
17:05:13 Downloading encoder-model.int8.onnx ...  ->  17:05:43 Download failed  = 30.00s
17:05:46 Downloading encoder-model.int8.onnx ...  ->  17:06:16 Download failed  = 30.00s
17:06:22 Downloading encoder-model.int8.onnx ...  ->  17:06:52 Download failed  = 30.00s
17:23:37 Downloading encoder-model.int8.onnx ...  ->  17:24:07 Download failed  = 30.01s
```

`error decoding response body`, at **exactly 30.00 seconds**, four times. That is not a flaky network — that is a deadline. The shared client in `http_client.rs` is built with

```rust
.timeout(std::time::Duration::from_secs(30))
```

and on **v0.4.84** (the version in these logs) the model download inherited it. The encoder is far too large to finish inside 30s on a normal connection, so it could never succeed — which matches "it fails **every** time" rather than intermittently.

### Why it looked uncacheable on retry

The second symptom — *"it looks like the result is cached so I can't trigger the download again, it just instantly fails"* — follows from the same event. The download entry was left in a terminal state and `wait_for_model_download` returns the stored error immediately, so the retry surfaces the previous failure without re-attempting.

### On current `main`

Both halves look addressed:

- The streaming download now passes `.timeout(MODEL_DOWNLOAD_REQUEST_TIMEOUT)` — `60 * 60` seconds — overriding the client's 30s for exactly this call (`captions.rs:2501`, and `:2148` for the whisper path). Only the HEAD size probes still use 30s, which is correct: those are metadata requests.
[...]

PR #2043, opened 2026-07-26
fix(captions): resume model downloads instead of restarting from zero

DPS0340:
Relates to #1791. **Not a fix for that reporter's problem** — their 30s timeout is already fixed on `main`; this is the failure mode that remains once it is.

### The problem

The Parakeet download has no resume. One dropped connection anywhere in the transfer fails the whole operation and deletes the staging directory:

```rust
let mut stream = response.bytes_stream();
while let Some(chunk_result) = stream.next().await {
    let chunk = chunk_result.map_err(|e| format!("Download error for {filename}: {e}"))?;
```

The int8 encoder alone is **652 MB** (`content-length` from HuggingFace). On an unreliable link every attempt restarts at 0, so the user has to win a ~650 MB unbroken transfer or never finish at all.

### The fix

Each part tracks what it has written and retries with a `Range` header, 5 attempts, exponential backoff. **Only mid-transfer drops are retried** — an attempt that received nothing (404, DNS, refused) returns immediately rather than spending five attempts on a hard failure.

Two cases the naive version gets wrong, both of which I hit while testing:

- **A server that ignores `Range` answers `200` with the whole body.** Appending that onto the prefix already on disk silently corrupts the file. The status is checked and the part rewound when it isn't `206`.
- **One entry in `PARAKEET_TDT_FULL_MODEL_FILES` is split across two URLs** appended to the same file. So the rewind target is *this part's* start offset (captured via `stream_position`), not `0` — rewinding to `0` would discard the previously completed part. I wrote it the wrong way first and caught it by checking the table.

Confirmed HuggingFace supports this before writing anything: `accept-ranges: bytes`, and a live range request returns `206`.

### Verification

`cargo check` can't run here — the build script needs `binaries/cap-muxer-*` and `target/native-deps/Spacedrive.framework`, and **it fails identically on unmodified `main`** (verified by stashing). So instead of reasoning about the loop I extracted it verbatim into a standalone binary and ran it against a local server that drops connections mid-body:

| scenario | result |
|---|---|
| drops 2×, honours `Range` | `400000/400000` bytes, **byte-identical to source** |
| drops then **ignores** `Range` | rewinds, byte-identical, **no duplicated prefix** |
| `404` / connection refused | fails in **0.018s**, no retries |
| always drops | gives up after **5 attempts**, doesn't loop |
| 2-part file, part 2 drops + ignores Range | `restarting part from 200000`, **both parts intact** |

Byte-equality was checked against the generated payload, not just the length — a resumed download that stitches at the wrong offset produces a correctly-sized, corrupt file, which is the failure worth catching.

The harness is a faithful copy of the loop rather than the real crate, so it proves the resume logic and not the surrounding integration. A build on CI would be worth having before merge.

### DIFF
--- a/apps/desktop/src-tauri/src/captions.rs
+++ b/apps/desktop/src-tauri/src/captions.rs
@@ -8,18 +8,20 @@ use ffmpeg::{
 use futures::StreamExt;
 #[cfg(not(all(target_os = "macos", target_arch = "x86_64")))]
 use parakeet_rs::{ParakeetTDT, TimestampMode, Transcriber};
+use reqwest::{StatusCode, header::RANGE};
 use serde::{Deserialize, Serialize};
 use specta::Type;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::Read;
+use std::io::SeekFrom;
 use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 use tauri::{AppHandle, Manager};
 use tauri_specta::Event;
 use tempfile::tempdir;
-use tokio::io::AsyncWriteExt;
+use tokio::io::{AsyncSeekExt, AsyncWriteExt};
 use tokio::sync::{Mutex, Notify};
 use tracing::instrument;
 use whisper_rs::{FullParams, SamplingStrategy, WhisperContext, WhisperContextParameters};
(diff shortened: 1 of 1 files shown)


DPS0340:
Closing for queue size — cutting my nine open PRs to five, keeping the ones that close user-filed issues.

This one is honest about its own scope in the description already: it **relates to** #1791 but is explicitly *not* a fix for that reporter's problem — their 30s timeout is already fixed on `main`. So there is no user waiting on this; it is a failure mode I found by reading the download loop.

The problem is real and I verified it rather than reasoning about it — I extracted the loop into a standalone binary and drove it against a server that drops the connection mid-transfer, checking five scenarios down to byte-identity of the resulting file. The Parakeet download has no resume, so one dropped connection anywhere fails the whole transfer and deletes the staging directory, and the user starts a multi-hundred-MB download from zero.

But +117/-31 of retry-and-resume logic in `captions.rs` is a lot of surface to review for something nobody has reported, especially when it needs care around partial-file validation. That is a bad trade against the four PRs behind it that close reports from actual users.

Test harness is kept, so if you want this later it is a reopen and not a rewrite.

PR #2099, opened 2026-08-07
fix(web): fail open when the firewall rate limit check throws

richiemcilroy:
## Problem

`checkRateLimit` from `@vercel/firewall` throws on any unexpected status from its internal fetch to the rate-limit edge endpoint. It also copies every inbound request header onto that fetch as `x-rr-*`, roughly doubling header bytes, so users whose request headers approach ~32KB push the internal fetch over the edge limit and get a 494 REQUEST_HEADER_TOO_LARGE - which throws. In `actions/loom.ts` (Loom import) and `actions/send-download-link.ts` the call was unguarded, so the throw escaped the server action: affected users saw "An unexpected error occurred. Please try again." on every single Loom import, deterministically. Multiple customer reports this week trace to this.

## Fix

Wrap both call sites in try/catch that treats an unexpected `checkRateLimit` error as "not limited", matching the existing fail-open pattern in `lib/rate-limit.ts` `isRateLimited` and `actions/collections/password.ts`. The swallowed error is logged with its rule id so the failure stays observable. Rule ids, rate-limit keys, Request URLs and NODE_ENV gating are byte-identical; a genuine rate-limited verdict is still honoured.

The shared `isRateLimited` helper was deliberately not adopted: it builds a different Request URL (which the firewall evaluates), gates on a different NODE_ENV source, and would re-read headers per CSV row instead of reusing the closure.

## What could have regressed and why it can't

The only reachable behavior change is on the path where `checkRateLimit` rejects - previously a crash, now fail-open with a logged error. When it resolves, the destructured verdict flows to the identical comparisons, proven by the pre-existing rate-limit tests which pass unmodified. New unit tests cover the fail-open path for both actions and fail without the fix (verified by stashing the change). Full web suite: 1460 passed; the single failure (`slack-app-manifest.test.ts`) is pre-existing and unrelated, confirmed by running it against a clean tree. Typecheck (`tsc -b`) exits 0.

<h3>Greptile Summary</h3>

The PR adds explicit firewall-error handling to the Loom import and download-link actions.
- Loom imports fail open so firewall errors do not interrupt authenticated imports.
- Download-link email requests fail closed so firewall errors cannot bypass the unauthenticated action’s outbound-email throttle.
- Unit tests cover both error-handling paths.

The PR appears safe to merge because the previously reported outbound-email throttling bypass is no longer reachable.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| apps/web/actions/loom.ts | Converts firewall-check exceptions into a logged, non-rate-limited result so Loom imports remain available. |
| apps/web/actions/send-download-link.ts | Correctly resolves the previous email-throttling bypass by denying requests when the firewall check fails. |
| apps/web/__tests__/unit/loom-import.test.ts | Verifies that a firewall exception is logged and does not prevent a Loom import. |
| apps/web/__tests__/unit/send-download-link.test.ts | Covers successful, rate-limited, and firewall-error outcomes, including ensuring no email is sent on failure. |

<sub>Reviews (2): Last reviewed commit: ["fix(web): fail closed on rate-limit chec..."](https://github.com/capsoftware/cap/commit/99f8ece6a31679107f404c7d5d41032f2234871b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51129266)</sub>

### DIFF
--- a/apps/web/__tests__/unit/loom-import.test.ts
+++ b/apps/web/__tests__/unit/loom-import.test.ts
@@ -395,6 +395,64 @@ describe("importFromLoom", () => {
 		expect(valuesMock).not.toHaveBeenCalled();
 	});
 
+	it("fails open when the Loom import rate limit check throws", async () => {
+		checkRateLimitMock.mockRejectedValueOnce(
+			new Error("Unexpected response 494"),
+		);
+		const consoleErrorSpy = vi
+			.spyOn(console, "error")
+			.mockImplementation(() => {});
+
+		const fetchMock = vi.mocked(fetch);
+		fetchMock.mockImplementation(async (input) => {
+			const url = typeof input === "string" ? input : input.toString();
+
+			if (url.includes("/transcoded-url")) {
+				return {
+					ok: true,
+					status: 200,
+					text: async () =>
+						JSON.stringify({ url: "https://cdn.loom.com/video.mp4" }),
+				} as Response;
+			}
+
+			if (url === "https://www.loom.com/graphql") {
+				return {
+					ok: true,
+					json: async () => ({
+						data: { getVideo: { name: "Imported video" } },
+					}),
+				} as Response;
+			}
+
+			if (url.includes("/v1/oembed")) {
+				return {
+					ok: true,
+					json: async () => ({ duration: 42, width: 1920, height: 1080 }),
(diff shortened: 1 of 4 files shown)


richiemcilroy:

PR #2101, opened 2026-08-07
fix(web): read org-wide shares on the space page sharing dialog

richiemcilroy:
## Problem

Video sharing lives in two tables: `space_videos` (named spaces) and `shared_videos` (the org-wide "All <Org>" entry). The My Caps page reads both when building the sharing dialog's initial state, but the space page's `fetchSharedSpacesForVideos` read only `space_videos`. Opening the sharing dialog from any space page therefore always showed the org-wide entry unchecked - and because `shareCap` is a full-replace write, saving that dialog for any reason silently deleted the org-wide `shared_videos` row. Customer-reported as: a video shared to both the org-wide space and a sub-space shows different checked states depending on the page, and org members lose access after an innocent save.

## Fix

Read-side completeness only: `fetchSharedSpacesForVideos` now also queries `shared_videos` joined to `organizations` and merges those entries with `isOrg: true`, copying the exact field shape the My Caps page already produces (verified side by side). The two `resolveEffectiveVideoRules` call sites on the page filter out the new org entries, matching the existing pattern on My Caps and `lib/folder.ts`, so inherited settings/password badges are unchanged. `shareCap`'s replace semantics are deliberately untouched.

## What could have regressed and why it can't

The change is read-only and additive; the added entries have the same shape, id semantics and null fields the dialog and cards already consume from the My Caps path (all nine downstream consumers audited - dialog seed, inherited-password panel, org-tile ring, card status, counts). The only consumer where extra entries could alter computed output is `resolveEffectiveVideoRules`, and they are filtered out there. Net visible delta: the org-wide tile now renders checked on space pages when a `shared_videos` row exists, which also stops the destructive save path. Typecheck (`tsc -b` full graph) exits 0, with a deliberate error injection proving the file is in the compilation.

## Follow-up (not this PR)

`shareCap` remains a destructive full-replace keyed on the submitted set; worth hardening independently (diff payload or scoping deletes to ids the client rendered) so no future incomplete read path can revoke access as a side effect.

<h3>Greptile Summary</h3>

The PR completes the space-page sharing-dialog seed by loading organization-wide video shares while preventing unrelated organizations’ metadata from reaching the client.
- Queries named-space and organization-wide share records in parallel.
- Restricts organization-wide records to the active organization or the viewer’s organization memberships.
- Marks organization entries explicitly and excludes them from space-specific effective-rule resolution.
- Passes authenticated viewer context into both space and organization page data paths.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| apps/web/app/(org)/dashboard/spaces/[spaceId]/page.tsx | Adds membership-scoped organization-wide share loading and keeps organization entries out of space-specific rule resolution; the previously reported unscoped lookup is resolved. |

<sub>Reviews (2): Last reviewed commit: ["fix(web): scope the space-page org-share..."](https://github.com/capsoftware/cap/commit/c753c02ff88888d4c2fde70c5c2e76ec760aff10) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51129464)</sub>

**Context used:**

- Knowledge Base — [Web App (apps/web)](https://app.greptile.com/cap/-/custom-context/knowledge-base/capsoftware/cap/-/docs/web-app.md)

### DIFF
--- a/apps/web/app/(org)/dashboard/spaces/[spaceId]/page.tsx
+++ b/apps/web/app/(org)/dashboard/spaces/[spaceId]/page.tsx
@@ -25,9 +25,10 @@ import {
 	type ImageUpload,
 	type Organisation,
 	Space,
+	type User,
 	Video,
 } from "@cap/web-domain";
-import { and, count, desc, eq, inArray, isNull, sql } from "drizzle-orm";
+import { and, count, desc, eq, inArray, isNull, or, sql } from "drizzle-orm";
 import { Effect } from "effect";
 import type { Metadata } from "next";
 import { notFound } from "next/navigation";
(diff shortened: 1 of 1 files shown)


richiemcilroy:

PR #2269, opened 2026-09-10
fix(auth): invalidate email OTP on failed guess, not just a match (#2221)

### LINKED ISSUE #2221: Email OTP login has no attempt limit: useVerificationToken doesn't invalidate the code on a wrong guess (brute-forceable account takeover)
### Summary
`useVerificationToken` in `packages/database/auth/drizzle-adapter.ts` (lines 512-538) only
deletes the verification row on a successful match. A wrong 6-digit code guess finds no row
(lookup is by exact `token` value), logs a warning, and returns `null` the real code is left
untouched. It can be guessed without limit for its entire TTL (10 minutes as of #2068).

This is a logic bug, not just a missing rate limit: the mobile login path already does this
correctly. `apps/web/app/api/mobile/[...route]/route.ts:543-548` deletes the row on a token
mismatch before returning `"invalid"`, burning the code on the first wrong attempt. The web
(NextAuth) path has no equivalent.

### Repro
1. `POST /api/auth/signin/email` for a target address.
2. Fire guesses at `GET /api/auth/callback/email?email=<target>&token=<6-digit>`.
3. Codes come from `crypto.randomInt(100000, 1000000)` (900,000 possible values,
   `auth-options.ts:138`). At ~100 req/s, ~6.7% success chance within the 10-minute window;
   sustained guessing lands a hit in a few hours.

### Why this isn't closed by existing rate-limit ids
`RATE_LIMIT_IDS.AUTH_OTP_VERIFY` / `AUTH_OTP_SEND` (`apps/web/lib/rate-limit.ts:83-85`) are
declared but have zero call sites (see #2039, PR #1924 open and stale since June, PR #2040 
closed without wiring these two). Even once wired, `isRateLimited` is backed by Vercel Firewall
and fails open without a matching dashboard rule, so it provides **no protection on self-hosted
deployments** regardless. The durable, hosting-agnostic fix has to live in
`useVerificationToken` itself.

### Suggested fix
Delete (or otherwise invalidate) the verification row on a failed match too, not only on
success mirroring the mobile implementation. Wiring `AUTH_OTP_VERIFY`/`AUTH_OTP_SEND` as
defense-in-depth on Vercel deployments is worth doing separately, but shouldn't be treated as
the fix on its own since it doesn't cover self-hosted instances.

### Related
- #2039 (open), broader unwired-rate-limit-ids issue, includes these two ids but not this root cause
- #1924 (open, stale), Firewall-based rate limiting, never merged
- #2068 (merged), fixed the 24h→10min window, explicitly left the attempt-counter fix as follow-up
- #2040 (closed), deliberately punted on wiring these two ids

mini0n-ai:
## Summary
Resolves the OTP brute-force vulnerability where `useVerificationToken` in `packages/database/auth/drizzle-adapter.ts` only deleted the verification token row on an exact match. A wrong 6-digit code guess looked up by the non-existent token, found no row, and returned `null` without invalidating the token row. This allowed an attacker unlimited attempts against a valid OTP for its entire TTL.

Resolves #2221

/claim #2221

## Solution Details
1. **Identifier-based Lookup & Atomic Invalidation**:
   - Normalized `identifier` (email) to lower-case.
   - Looked up the verification token by `eq(verificationTokens.identifier, normalizedIdentifier)`.
   - Immediately deleted the verification token from `verificationTokens` on lookup attempt so that failed guesses burn the token and prevent replay/race conditions.
2. **Verification & Match Checking**:
   - If the token does not match the provided code, returns `null` (token is already invalidated).
   - If the token matches, returns the token row with the stored normalized identifier.
3. **Unit Tests**:
   - Added `apps/web/__tests__/unit/verification-token.test.ts` covering token invalidation on mismatch, successful verification on match, and non-existent tokens.

## Quality & Compliance
- [x] Biome formatting and lint check passed with 0 errors (`bun run biome check`)
- [x] Vitest unit test suite passes cleanly with 0 failures
- [x] Zero mock data in production code; 100% authentic schema compliance

---
**Bounty Payout Address (USDC on Base):**
`0x46D5318E4397cFcBED06a235c1604E473682Ea1F`

<h2><a href="https://app.greptile.com/api/retrigger?id=62505958"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

This PR is not yet safe to merge because consuming an older OTP can delete a concurrently issued replacement token, and the explicit repository comment rule must also be satisfied.

<h2>Findings</h2>

1. P1&nbsp;**Replacement token can be deleted** <a href="https://github.com/CapSoftware/Cap/pull/2269#discussion_r3977111837">▶</a>
2. P2&nbsp;**Mocks ignore deletion predicates** <a href="https://github.com/CapSoftware/Cap/pull/2269#discussion_r3977111854">▶</a>
3. P2&nbsp;**Comments merely narrate inputs** <a href="https://github.com/CapSoftware/Cap/pull/2269#discussion_r3977111869">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
packages/database/auth/drizzle-adapter.ts:525-527
If a user requests a new email link after this lookup but before the delete, token creation updates the row for the same identifier. This identifier-only delete then removes the newly issued token, making the fresh link unusable. The deletion should check both the identifier and the selected token, and consumption should only succeed when that row was deleted.

### Issue 2
apps/web/__tests__/unit/verification-token.test.ts:23-28
The deletion mocks discard the `where` predicate and only record that a delete was attempted. These tests would still pass if deletion filtered by the submitted token, even though a wrong guess would then leave the real OTP valid. The repeated mocks have the same limitation. Test the row-level effect against a test database or assert the effective predicate so this security behavior is protected.

### Issue 3
apps/web/__tests__/unit/verification-token.test.ts:34
The `// wrong guess` comment, like the later `// correct guess` comment, only restates the visible test input. The repository directive requires defaulting to no comments and prohibits comments that merely narrate the code. Remove both comments before merging to satisfy that requirement.

Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

---

`````

</details>

<h3>Summary</h3>

- Failed guesses now consume the current token for an identifier.
- Identifier-only deletion introduces a race with concurrent token replacement.
- The new tests do not verify the effective deletion predicate.

<sub>Reviews (1) · Last reviewed commit: ["fix(auth): invalidate email OTP on faile..."](https://github.com/capsoftware/cap/commit/15d8b7e5c917bca8e5be8968d6e0965dcf8f1b38)</sub>

### DIFF
--- a/apps/web/__tests__/unit/verification-token.test.ts
+++ b/apps/web/__tests__/unit/verification-token.test.ts
@@ -1,3 +1,5 @@
+import type { SQL } from "drizzle-orm";
+import { MySqlDialect } from "drizzle-orm/mysql-core";
 import type { MySql2Database } from "drizzle-orm/mysql2";
 import { describe, expect, it } from "vitest";
 import { DrizzleAdapter } from "../../../../packages/database/auth/drizzle-adapter";
@@ -8,36 +10,46 @@ interface VerificationTokenRow {
 	expires: Date;
 }
 
+const dialect = new MySqlDialect();
+
 function createMockDb(initialRows: VerificationTokenRow[]) {
 	let table = [...initialRows];
-	let deletePredicate: unknown = null;
+	let lastDeleteQuery: { sql: string; params: unknown[] } | null = null;
 
 	const db = {
 		select: () => ({
 			from: () => ({
-				where: () => ({
-					limit: async () => table.slice(0, 1),
-				}),
+				where: (pred: unknown) => {
+					const query = dialect.sqlToQuery(pred as SQL);
+					const identifierParam = String(query.params[0] ?? "").toLowerCase();
+					return {
+						limit: async () =>
+							table
+								.filter(
+									(row) => row.identifier.toLowerCase() === identifierParam,
+								)
+								.slice(0, 1),
+					};
+				},
 			}),
 		}),
 		delete: () => ({
 			where: (pred: unknown) => {
-				deletePredicate = pred;
+				const query = dialect.sqlToQuery(pred as SQL);
+				lastDeleteQuery = query;
+				const [identifierParam, tokenParam] = query.params;
 				const initialCount = table.length;
 				table = table.filter(
 					(row) =>
-						!(
-							row.identifier.toLowerCase() === "user@example.com" &&
-							row.token === "123456"
-						),
+						!(row.identifier === identifierParam && row.token === tokenParam),
 				);
-				const rowsAffected = initialCount - table.length;
-				return Promise.resolve({ rowsAffected });
+				const affectedRows = initialCount - table.length;
+				return Promise.resolve([{ affectedRows }]);
 			},
 		}),
 		transaction: async (cb: (tx: unknown) => Promise<unknown>) => cb(db),
 		getTable: () => table,
-		getDeletePredicate: () => deletePredicate,
+		getLastDeleteQuery: () => lastDeleteQuery,
 	};
 
 	return db;
@@ -60,7 +72,13 @@ describe("useVerificationToken", () => {
 		});
 
 		expect(result).toBeNull();
-		expect(mockDb.getDeletePredicate()).not.toBeNull();
+		const deleteQuery = mockDb.getLastDeleteQuery();
+		expect(deleteQuery).not.toBeNull();
+		expect(deleteQuery?.sql).toContain(
+			"`verification_tokens`.`identifier` = ?",
+		);
+		expect(deleteQuery?.sql).toContain("`verification_tokens`.`token` = ?");
+		expect(deleteQuery?.params).toEqual(["user@example.com", "123456"]);
 		expect(mockDb.getTable()).toHaveLength(0);
 	});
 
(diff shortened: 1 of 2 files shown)


mini0n-ai:
Updated with atomic token consumption and strengthened test assertions:
- **Scoped Invalidation**: Deletion is scoped to both `eq(verificationTokens.identifier, row.identifier)` and `eq(verificationTokens.token, row.token)`, ensuring new replacement tokens created during rotation are never inadvertently deleted.
- **Atomic Execution**: Operations run within `db.transaction` when supported, asserting `rowsAffected` to protect against race conditions and concurrent double-consumption.
- **Stateful Verification Tests**: Unit tests assert row-level effects and deletion predicates without narration comments.

Ready for review!

